PR #101, opened 2026-07-22
Add TDA blog post: multi-scale clustering via persistent homology

project-delphi:
## Summary
- New post: *Topological Data Analysis: Finding Clusters, Loops, and Voids Across Scales*
- Builds persistent homology up from first principles — simplices, the Vietoris-Rips filtration, and a hand-rolled boundary-matrix computation of Betti numbers on a 4-point "hollow square" example, cross-checked against `ripser`
- Shows that β₀ persistence is exactly single-linkage clustering, and frames higher `H_k` as the same cycles-modulo-boundaries computation one dimension up ("clustering, expanded across dimensions")
- Covers how TDA differs from k-means/DBSCAN/GMM in its multi-scale, assumption-free approach, plus a toy example using persistence diagrams as features for a downstream classifier
- All code cells execute live (via a dedicated local Jupyter kernel with numpy/scipy/sklearn/ripser/persim) and render real plots/output, rather than being illustrative-only snippets

## Test plan
- [x] `quarto render posts/topological-data-analysis-clustering/index.qmd` succeeds with all 6 cells executing
- [x] Hand-rolled Betti number computation matches `ripser` output on the hollow-square example
- [x] H0-persistence-equals-single-linkage check prints `True`
- [x] Rendered `docs/` output reviewed visually (filtration plot, persistence diagram plot render correctly)


project-delphi (commented):
Self-review notes:

- The hand-rolled boundary-matrix Betti number computation (hollow-square example) matches `ripser`'s output exactly (β₀: 4→1, β₁: 0→1→0 at ε=1 and ε=√2), so the first-principles section is verified against the standard tool rather than just asserted.
- H0-persistence-equals-single-linkage-clustering claim is checked in code (`np.allclose` over sorted death times vs. merge heights), not just stated.
- Code cells execute live via a dedicated Jupyter kernel (`tda-blog`) — flagging that this kernel/venv is local-only and not part of the repo. If this post needs to be re-rendered from a clean clone later, that environment (numpy/scipy/sklearn/ripser/persim) will need to be recreated; happy to fold it into `pyproject.toml` + the Makefile's `.venv` target in a follow-up if we want that reproducible in-repo.
- No new site infra changes — this is a content-only PR (one new post + regenerated `docs/index.html`, `listings.json`, `search.json` from `quarto render`).

PR #107, opened 2026-07-23
fix(posts): sharpen the agent definition in LangGraph vs LlamaIndex post

project-delphi:
## Summary
- Rewrites the "What an LLM Agent Actually Is" section of `posts/langgraph-vs-llamaindex/`: the previous `Agent` sketch hand-waved `interpret`/`update_state` without showing the actual anatomy of an agent.
- New definition is structural: the `LLM` is a stateless function from a message sequence to one new message (final answer or tool-call request), and the `Agent` is a class holding a **message sequence** (the transcript as the model's only memory), **task state**, a **tools registry**, and a **stop condition + max-turn cap** — with a `run` method that is a deterministic loop around the single stochastic `llm.chat` call.
- The closing paragraph now ties the post's later reply-cap requirement back to `max_turns` surfacing as a business rule, so the primer connects to the main comparison instead of being a detour.
- Rendered output (`docs/`) included, per this repo's no-CI/render-locally convention.

## Test plan
- [x] `quarto render posts/langgraph-vs-llamaindex/` — renders clean, no errors
- [x] Previewed in-browser (served `docs/` statically): rewritten section renders correctly, both new code folds expand with proper highlighting, rest of the post unaffected


project-delphi (commented):
## Review Summary

**Critical Issues:** 0
**Important Issues:** 0
**Suggestions:** 0

### Overview
Prose + illustrative-code rewrite of the "What an LLM Agent Actually Is" primer in the LangGraph vs LlamaIndex post. The previous `Agent` sketch deferred all the interesting parts to undefined `interpret`/`update_state` methods; the new version shows the real anatomy — a `Message` dataclass, an `LLM` as a stateless messages-in/one-message-out function, and an `Agent` class holding the message transcript, task state, a tools registry, and a max-turns stop condition, with a deterministic `run` loop around the single stochastic call.

### Verification
- Rendered HTML (`docs/posts/langgraph-vs-llamaindex/index.html`) confirmed in sync with the qmd: new loop present, old `def step` sketch fully removed.
- The sketch is internally consistent: `is_done` is defined (unlike the old version's hand-waved methods), the tool-dispatch line matches the `Message.tool_call` shape declared above it, and the loop provably terminates via `range(self.max_turns)` with a fail-safe `escalate()` — which doubles as a deliberate hook into the post's v2 escalation section.
- Deliberate simplifications are appropriate for an `eval: false` blog sketch: no imports shown (consistent with every other excerpt in this post) and `self.state` unused in the loop (the prose explicitly frames it as "task state beyond the transcript," which is precisely the axis the rest of the post explores per framework).
- `docs/index.html` / `docs/search.json` changes are just the regenerated listing metadata from the render — expected under this repo's render-locally convention.

### Strengths
- The closing tie-back — the post's reply-cap requirement is "`max_turns` showing up as a business rule" — makes the primer load-bearing for the main comparison instead of a detour.
- "The only nondeterminism anywhere is inside `llm.chat`; everything else is ordinary code you could unit-test with a mocked model" is a genuinely useful framing that most agent-framework intros omit.

No issues found. Approving for merge.

PR #108, opened 2026-07-23
feat(posts): add TDA filtered-simplicial-complex post

project-delphi:
## Summary

- New post: **"What TDA Actually Builds: A Filtered Simplicial Complex, Not a Fancy Graph"** — a first-principles walkthrough correcting graph-centric intuitions about TDA (edges vs filled 2-simplices, hypergraph intuition vs the closure property, filtration vs persistence computation, scale vs time).
- One fixed 8-point running example (seed 42) carries every section: distance matrix, nested threshold graphs, Rips complexes with filled triangles, ripser persistence diagram/barcode, and Betti numbers cross-checked against networkx component counts.
- Follows repo conventions: dedicated `.venv-tda-filtered` + `tda-filtered-blog` kernel pinned in frontmatter, house-style purple cover card, rendered `docs/` committed alongside the source.

## Test plan

- [x] `quarto render posts/tda-filtered-simplicial-complexes/index.qmd` succeeds; all 11 cells execute in the dedicated kernel
- [x] All displayed numbers (distances, birth/death values, Betti numbers, simplex counts) computed live from the running example; runtime `assert` cross-checks β₀ against networkx
- [x] Every generated figure visually inspected for readable labels/legends and non-overlapping panels
- [x] Built page previewed locally from `docs/`; homepage listing shows the new card/cover

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #109, opened 2026-07-23
fix(posts): apply review polish to TDA complexes post

project-delphi:
## Summary

- Applies the four non-blocking suggestions from PR #108's review:
  - `draw_complex` now computes distances locally instead of depending on the global `D` defined two cells later (also removes the `eps > 0` guards and the unusual inline-conditional loop idiom that coupling forced)
  - `simplex_counts` no longer computes 4-simplex counts the table never displays
  - the reproducibility callout now documents the `QUARTO_PYTHON` env var Quarto needs to discover the `tda-filtered-blog` kernel
- No content or behavior change: re-rendered figure PNGs are byte-identical; only the qmd source, its rebuilt HTML, and the site listing/search index changed.

## Test plan

- [x] `quarto render posts/tda-filtered-simplicial-complexes/index.qmd` succeeds; all cells execute (including the β₀ cross-check `assert`)
- [x] Confirmed no figure files changed in `git status` after re-render (byte-identical output)


project-delphi (commented):
## Code Review

### Overview
Follow-up applying the four non-blocking suggestions from PR #108's review: `draw_complex` computes its own distance matrix instead of reaching for the global `D` defined two cells later, the `eps > 0` guards and inline-conditional loop idiom that coupling forced are gone, `simplex_counts` stops computing an undisplayed 4-simplex column, and the reproducibility callout documents the `QUARTO_PYTHON` env var. Rebuilt `docs/` HTML and listing/search index are committed per repo convention.

### Critical Issues
None found.

### Important Issues
None found.

### Suggestions
- `draw_complex` now recomputes `pdist` on each of its ~9 calls — trivially cheap for 8 points and arguably clearer, so fine as is; just noting it's a deliberate trade of a redundant computation for decoupling.

### Strengths
- Behavior-preserving refactor with strong evidence: every figure PNG is byte-identical after re-render, so only source, HTML, and index files appear in the diff.
- The ε = 0 edge case that motivated the original guards is genuinely covered without them (pair loops visit only distinct points, all at positive distance).
- The `QUARTO_PYTHON` documentation quotes the actual error message a future renderer would see, making it searchable.

### Verdict
No blocking issues — safe to merge.

PR #111, opened 2026-07-23
feat(posts): SVM maximum-margin convex optimization post

project-delphi:
## Summary
- New post `posts/svm-max-margin-convex-optimization/`: builds the linear SVM from first principles as explicit CVXPY quadratic programs (hard margin, then soft margin with slack/C sweep) on small hand-checkable 2D datasets, with sklearn only as a final cross-check.
- Executed through a dedicated venv/kernel (`svm-margin-blog`) per the repo's per-post isolation convention; rendered output committed under `docs/` (post page, figures, refreshed listing/search/index).
- CLAUDE.md: documents the hard rule of never installing into or executing with the system Python — always go through the post's venv (activate it or call `.venv-<slug>/bin/python -m ...` directly), and verify the registered kernel's `argv` points at the venv interpreter.
- `.gitignore`: adds `.claude/settings.local.json`.

## Test plan
- [x] `quarto render posts/svm-max-margin-convex-optimization/index.qmd` completed with all in-notebook assertions passing (solver values checked against analytic derivations)
- [x] Rendered HTML and all five figures present under `docs/posts/svm-max-margin-convex-optimization/`
- [x] Homepage listing and search index refreshed by the targeted render


project-delphi (commented):
## Code review (Claude Code)

### Overview
Adds a new first-principles post building the linear SVM as explicit CVXPY quadratic programs — hard margin on a hand-constructed separable dataset, convexity/uniqueness analysis (including a genuinely degenerate one-class case and the flat-intercept soft-margin case), Lagrangian/dual verification, and a soft-margin C sweep on non-separable data — with sklearn as an independent cross-check. Also ships the rendered `docs/` output, a CLAUDE.md rule against using system Python, and a `.gitignore` entry for Claude Code local settings.

### Critical Issues
None found.

### Important Issues
None found.

### Correctness (verified independently)
- Hand-checked the analytic hard-margin solution: perpendicular bisector of (1,1)/(3,3) gives w*=(0.5,0.5), b*=−2; verified all 8 non-SV constraints are strictly slack (min inactive functional margin 1.1), so `sv_mask_hard.sum() == 2` is robust.
- Dual claims check out: α=0.25 per SV reproduces w*, Σαᵢyᵢ=0, and dual value 0.25 equals primal ½‖w*‖².
- Flat-b derivation is correct: with C=0.25 the two hinge terms cancel b exactly on |b|≤0.5, objective floor 0.375, boundary x₁=−2b ∈ [−1,1], consistent with Burges & Crisp.
- Candidate-line margins in the prose (1.414 vs 0.86) match hand computation.
- The post's in-notebook `assert`s mean the committed HTML cannot drift from the prose — and the rendered `docs/` output confirms all assertions passed.

### Suggestions (non-blocking)
- **Naming drift vs CLAUDE.md convention**: CLAUDE.md prescribes `.venv-<slug>`; this post uses `.venv-svm-margin` with kernel `svm-margin-blog` while the slug is `svm-max-margin-convex-optimization`. Works fine, but a future fresh-clone rebuild has to discover the mapping from the frontmatter. Consider noting the kernel↔venv pairing convention doesn't require the full slug, or align names next time.
- **Implicit venv ignoring**: the repo-root `.gitignore` only ignores `.venv` exactly; the per-post venvs stay untracked only because Python's `venv` module writes a `*` `.gitignore` inside each one. Adding `.venv-*/` to the root `.gitignore` would make this explicit and robust to venvs created by other tools.
- Dataset B's infeasibility assertion depends on NumPy's seeded `default_rng` stream; NumPy guarantees stream stability across versions, so this is fine — just worth knowing it's load-bearing.
- `hard.solve(solver=cp.CLARABEL)` at the end of the uniqueness cell restores solver state that no later cell actually reads (downstream uses the captured `w_hard`/`b_hard`); harmless.

### Strengths
- Every numeric claim in prose is asserted in code before the post can render — an unusually strong guard against prose/number drift.
- Independent verification at three levels: analytic derivation, two different solver algorithms (interior-point vs first-order), and a separate implementation (sklearn SMO on the dual).
- Honest treatment of the sharp edges (infeasibility, non-uniqueness, solver tolerances) that most SVM intros skip.
- Follows house conventions: frontmatter shape, cover image, dedicated kernel pinning, `freeze: auto`-compatible targeted render (only this post's docs plus listing/search touched).
- Accessible figures: class encoded by shape and color, shared axes across the C-sweep panels.

**Verdict: approve — no blocking issues.**

PR #117, opened 2026-07-26
feat(posts): add "The Poor Person's Bayesian" post

project-delphi:
## Summary

- New post **The Poor Person's Bayesian** (`posts/poor-persons-bayesian/`) working through Brian Ripley's framing — *the bootstrap moves the data; Bayes moves the parameters* — and locating the point where the analogy stops holding: the Dirichlet/multinomial correspondence between Rubin's Bayesian bootstrap and Efron's, BCa, and the small-`n` regimes where only a proper prior has anywhere left to put mass.
- First **knitr/R** post in the blog (`engine: knitr`, no Jupyter kernel). `cover.png` is written *during* execution by the `fig-law` chunk, so the frontmatter names it explicitly under `resources:` — without that the listing thumbnail and social card 404.
- Post-scoped `styles.scss` layered over the site's `zephyr` theme; brings in two new `docs/site_libs/` CSS bundles.
- `.gitignore`: ignore `posts/**/index_cache/` and `posts/**/index_files/` — knitr writes build artifacts into the post source dir, and the published copies already live under `docs/`.
- `docs/` rebuilt and committed per repo convention (no CI renders this site).

## Test plan

- [x] `quarto render posts/poor-persons-bayesian/index.qmd` — all R chunks execute clean, 9 figures produced.
- [x] `quarto render index.qmd` — homepage listing regenerated; the post now appears as the top row of `docs/index.html`, and is present in `docs/listings.json` / `docs/search.json`.
- [x] `cover.png` present in both `posts/` and `docs/posts/` (it is render-generated, so verified rather than assumed).
- [x] Confirmed `index_cache/` and `index_files/` under `posts/` are excluded from the commit while the `docs/` copies are included.
- [ ] Not verified: rendered page in a browser / social-card preview against the live GitHub Pages URL.


project-delphi (commented):
## Review — PR #117

Reviewed the diff (`gh pr diff 117`) plus the rendered `docs/` output, and re-ran the key numbers in R to check the claims. This is a strong post; the findings below are about a few load-bearing assertions and about what happens the next time someone re-renders it.

### Critical Issues

**1. `posts/poor-persons-bayesian/index.qmd:405` — the `(n+1)/n` attribution is quantitatively wrong, and the post's own table contradicts it.**

> "Its weights are smoother and — **by exactly the $(n+1)/n$ of @eq-weightvar** — its resampling distribution is slightly tighter than Efron's, so its intervals are slightly shorter."

Re-running the `fig-law` setup (seed 1979, `B = 10000`):

```
percentile interval length  0.4965
BB interval length          0.4449   ratio 0.896
sd(Efron)  0.1317   sd(BB)  0.1178   ratio 0.894
sqrt(n/(n+1)) at n = 15              = 0.968
```

And in @tbl-coverage the mean lengths are BB 0.382 vs percentile 0.469 — a ratio of **0.815**.

So the weight-variance factor predicts a 3.2% shrinkage; the observed shrinkage is 10.6% on the single sample and 18.5% averaged over the 500 subsamples. `(n+1)/n` accounts for under a third of the effect at best, not "exactly" it. Correlation is a nonlinear functional, so the weight-variance ratio only transfers at first order; the rest of the tightening comes from Efron's discreteness throwing occasional near-degenerate resamples that fatten the tails.

The section's *thesis* — smoothness is not calibration — is untouched and correct. It's the stated mechanism that overclaims, in a paragraph explicitly flagged as "this post's thesis arriving two sections early", so it's worth getting right.

### Important Issues

**2. `index.qmd:331`, `index.qmd:885` — `#| cache: true` with no `dependson` / `autodep`.**

`fig-law82` consumes `dens_df` and `n_law` from the uncached `fig-law` chunk, plus `rdirichlet1` / `bb_cor_mat` / `fmt` / `bm` / the palette from `setup`. knitr's `autodep` defaults to `FALSE`, so editing any upstream chunk (seed, `B`, palette, helper) leaves the cached figure, `covtab`, and every inline number drawn from it silently stale — while the surrounding prose compares those cached numbers directly against freshly-computed ones. `fig-nsweep` has the same exposure to `setup`. Suggest `#| dependson: [setup, fig-law]` on `fig-law82`, or `knitr::opts_chunk$set(autodep = TRUE)` in `setup`.

**3. 59 `size = ` call sites will emit deprecation warnings into the published page on the next re-render.**

`sessionInfo()` in the committed output shows **R 4.0.5 / ggplot2 3.3.5**. From ggplot2 3.4.0, `size` on line-based geoms (`geom_segment`, `geom_line`, `geom_path`, `geom_hline`, `geom_vline`, `stat_function`) and on `element_line()` / `element_rect()` is deprecated in favour of `linewidth`. Only the `setup` chunk sets `warning: false`, and neither `posts/_metadata.yml` nor `_quarto.yml` sets a global execute option — so on a current toolchain those warnings render as visible output blocks throughout the post. Either add `execute: {warning: false}` to the frontmatter or migrate to `linewidth`.

**4. `index.qmd:342` — the truth check's estimand doesn't quite match what the intervals target.**

`law82[sample(nrow(law82), n_law), ]` is a simple random sample **without** replacement: 15 of 82, an 18% sampling fraction. The resulting "actual sampling distribution" therefore carries a finite-population correction of $\sqrt{1 - 15/82} \approx 0.90$ on the sd, whereas all four interval constructions target the iid-from-infinite-population law. The direction is favourable — a narrower truth means intervals calibrated for iid should *over*-cover, so the reported under-coverage is conservative — but @tbl-coverage is presented as clean empirical coverage and deserves a caveat sentence.

### Suggestions

- **Document-global `B` and `n`.** Nearly every chunk rewrites both, and inline text depends on evaluation order — `r bm(B)` at line 495 happens to read 30,000 from `fig-ties`. Correct today, but it interacts badly with issue 2. Per-chunk names (`B_ties`, `B_law`) would make the inline references self-evidently right.
- **`base_family = "Palatino"` (line 100)** has no fallback; a machine without it emits font warnings on every device call and shifts all nine figures.
- **Hardcoded constants in prose:** `(1 - 1/25)^{25}` at line 879 where `n_u <- 25` exists, and `24` at line 577 where it's `2 * 12`. Inline `r` would keep them honest if the setup changes.
- `ggsave("cover.png")` inside `fig-law` writes into the source tree on every render, so `git status` shows churn after each rebuild. Intentional and documented in the README — noting it only so it isn't mistaken for an accident later.

### Strengths

- **The truth check is the thing most posts skip.** Having `law82` and actually reporting coverage that doesn't flatter the author's preferred method is the strongest move in the post.
- The Haldane–Anscombe ≡ Jeffreys callout, and the refusal to leave it as an aside.
- The even-handedness turn at line 682 (the advantage *flips* on genuinely discrete data) — pre-empts the obvious objection instead of absorbing it.
- `fig-sandwich` demonstrates "orientation, not scale" and then checks the principal-axis angle numerically rather than asking the reader to eyeball the ellipses.
- Every figure carries real `fig-alt`; the `.gitignore` patterns are correctly anchored so `posts/**/index_files/` is ignored while the needed `docs/posts/**/index_files/` are still committed.
- References verified against publisher records, with image provenance in `images/attribution.txt`.

*Review performed inline against the PR diff and rendered output, with the numerical claims re-executed in R.*

project-delphi (commented):
## Re-review after fixes — c3ae780

All Critical and Important findings addressed; re-rendered clean (2m34s, no warnings, no stderr blocks in the output).

**1. Critical — `(n+1)/n` attribution.** Fixed. The paragraph now measures the gap instead of asserting the mechanism, with every number computed inline from live objects rather than typed in:

> That factor predicts a spread ratio of $\sqrt{n/(n+1)} = 0.968$ at $n = 15$. What actually happens is a standard-deviation ratio of **0.894** on this sample, and a mean interval length across the 500 subsamples of **0.814** of the percentile interval's. The weight variance accounts for well under half of the shrinkage it usually gets credited with.

The remainder is attributed to nonlinearity (≈9.7 distinct schools per Efron resample vs all 15 reweighted), with the tail asymmetry stated as observed. The section's thesis — smoothness is not calibration — is unchanged, and the closing line now makes the fixed version load-bearing: *shorter and worse-covered is exactly what an uncorrected interval on a smoother resampling law looks like.*

**2. Important — stale chunk cache.** Fixed, though not the way I first suggested: `dependson` turns out to be unusable here, since knitr rejects it against uncached chunks (`code chunks must not depend on the uncached chunk "setup"`). Caching `setup` and `fig-law` to make it legal would have spread the hazard rather than removed it. `cache: true` is instead dropped from both `fig-law82` and `fig-nsweep`, with a comment recording why — `freeze: auto` in `posts/_metadata.yml` already caches at document level and invalidates on any source change, which is the correct-by-construction behaviour. Cost is a 2m34s cold render; the coverage table reproduced bit-identically (0.918 / 0.932 / 0.868 / 0.934), confirming the seeds were doing their job.

**3. Important — ggplot2 deprecations.** `execute: warning: false` added to the frontmatter, with a comment naming the ggplot2 3.3.5 → 3.4 `size`/`linewidth` transition as the reason. Migrating the 59 call sites would break the current build, since `linewidth` doesn't exist in 3.3.5.

**4. Important — finite-population correction.** Caveat paragraph added before @tbl-coverage, stating the 18% sampling fraction, the $\sqrt{1 - 15/82} \approx 0.90$ correction, and that the mismatch runs conservative — so the reported under-coverage is if anything understated.

**Suggestions** — hardcoded `25` and `24` in prose replaced with inline `r n_u` / `r n_s`. Left open as genuinely minor: document-global `B`/`n`, the missing `Palatino` fallback, and `ggsave` writing into the source tree (documented in the README).

**Verdict: no Critical issues outstanding. Clear to merge.**

PR #118, opened 2026-07-27
feat(posts): add "Why the Bayesian Bootstrap?" first-principles post

project-delphi:
## Summary

A long-form post deriving the Bayesian bootstrap from first principles and setting it against Efron's, on S&P 500 / IBM returns and MNIST.

- **Derived, not cited.** The Dirichlet falls out of multinomial likelihood × prior with the exponents incrementing; the gamma representation is done with the actual Jacobian; the median's exact rank law comes from uniform spacings and lands on the classical sign-test interval. The one result not derived (Lo's asymptotic equivalence) is flagged as such and its consequence checked numerically.
- **Self-checking build.** Every closed form stated in the prose is also computed and `assert`ed against simulation, with tolerances set by Monte-Carlo standard error. Exact enumerations where they're cheap: all 126 Efron patterns at n=5, the Poisson-conditioning identity at n=3,4,5 to machine precision, the 35 equiprobable Dirichlet-multinomial patterns at n=4. The page cannot render while the algebra and the arithmetic disagree.
- **Honest about ties.** At n=59 the two methods differ in the third decimal, which is inside Monte-Carlo noise at B=200,000 — the post says so rather than over-reading it, and spends its length on the two places the choice genuinely bites (rare events, extrema).
- **Widgets are dependency-free vanilla JS**, not Observable. OJS does not run under the pinned Quarto 1.6.40: the bundled runtime wedges its scheduler with all 47 stdlib variables queued, reproducible on a document containing only `x = 41 + 1`. The rewrite also dropped the CDN dependency, so the widgets work from `file://`.
- Ships light-only by design; `color-scheme: light` opts out of Chrome's auto-dark, which would otherwise invert the page while leaving figures light.

Follows the repo conventions in `CLAUDE.md`: self-contained `posts/<slug>/`, dedicated `.venv-bayesian-bootstrap` + `bayesian-bootstrap-blog` kernel pinned in frontmatter, post-scoped `theme.scss` layered over `zephyr`, `docs/` rendered and committed.

## Test plan

- [x] Renders clean via `QUARTO_PYTHON=... quarto render posts/bayesian-bootstrap/index.qmd`; all 62 cells execute, all assertions pass
- [x] Ground truth independently verified against the live data before the prose was written — IBM/S&P sample descriptions, median sign-test interval (order stats 23 and 37), tail atom 0.3647, two-sample comparison, Beta accuracy identity all reproduce
- [x] Figure layout checker (`src/layout_qa.py`, a build gate): **10 figures, 0 collisions**
- [x] All 7 widgets exercised in Chrome; readouts match the verified values (W3 ratio 0.9608/0.9989, W6 preset 36.5%, W5 Beta(1,58) 95% [0.00044, 0.0616])
- [x] Degenerate cases special-cased — W4 "max" draws the spike labelled "posterior is a point mass at 6.171", not an empty histogram
- [x] 0 broken math spans (inline interpolation moved out of `$…$`, where pandoc escaping rendered `0\.9916` literally)
- [x] No horizontal overflow at 375 px; twin blocks stack; 18 tables scroll in their own boxes
- [x] WCAG AA on the paper ground: ink 15.33, plum 7.01, petrol 6.65, dark ochre 5.07, muted 4.53
- [x] Widget data 21.6 KB of a 1 MB budget, no precomputed replicate arrays shipped
- [x] Homepage listing regenerated; new post first, thumbnails load
- [x] `make help / data / widgets` verified; `data/` and the venv confirmed gitignored

## Notes for the reviewer

- **Render is ~110 s, not the <60 s the brief targeted.** Quarto only consults `freeze` on *project* renders — single-file renders always re-execute. Confirmed in a controlled test (10.5 s → 10.3 s single-file vs 1.8 s project). The freeze cache is written with a stable hash, so the publishing path does hit it.
- **Long window is n=437, not the 438 in the brief** — verified against the live Shiller series; shifts that row's standard errors in the fourth decimal. Computed values win; recorded in the README.
- Two upstream Quarto bugs found and written up in the README, including one worth knowing generally: a multi-statement `{ojs}` cell emits ids `ojs-cell-1-1…` while registering as `ojs-cell-1`, so lookup returns null and the failure aborts every later cell.


project-delphi (commented):
## Review — PR #118

### Overview

Adds a self-contained long-form post under `posts/bayesian-bootstrap/`: `index.qmd` plus a five-module Python package (`bootstrap.py`, `data.py`, `mnist_experiment.py`, `export_widget_data.py`, `layout_qa.py`), a post-scoped `theme.scss`, seven dependency-free JS widgets, and the rendered `docs/` output. 16 source files; the other 20 changed files are generated output.

The design is unusual in a good way: the post is *executable and self-checking*. Every closed form stated in the prose is recomputed and `assert`ed against simulation at build time, and a figure-layout checker is wired in as a build gate, so the page cannot render while the algebra and the arithmetic disagree.

---

### Critical Issues

**None.** Nothing here blocks merge.

---

### Important Issues

**1. Dead code in `index.qmd:1596-1597`**

```python
n_metrics = 4000
Wd = bs.dirichlet_weights(n_te, n_metrics, rng, ) if False else None
```

`Wd` is never read, the `if False` branch never executes, `n_metrics` exists only to feed the dead call, and there is a stray trailing comma in the argument list. Confirmed by grep: `n_metrics` appears on exactly these two lines and nowhere else.

This matters more than usual here because the post's premise is that the source is the argument and is auditable. `if False` in a shipped artifact undermines that for anyone who opens the file. Delete both lines.

**2. `correct_rle` is shipped but never consumed**

`export_widget_data.py:179` writes a run-length encoding of the 10,000-element correctness vector into `mnist.json`, and `_rle` (`:140`) exists solely to produce it. No widget decodes it — W7 reads only `per_class`, `n` and `k`. Confirmed by grep across `widgets.js` and `src/`.

The encoder is correct (it emits a leading zero-length run when the vector starts `False`, which is the right convention), so this is dead weight rather than a bug. Either wire it up or drop both the field and the function.

---

### Suggestions

**3. `data.py:114` — the incomplete-month heuristic is brittle.** `if last_day.day < 20: drop the last month` works for this dataset (which ends 2018-02-07) but would silently keep a partial month for a series ending on, say, the 25th. Comparing `last_day` against the month's expected last business day would be robust to a future data refresh.

**4. `metric_boot` (`index.qmd:1599`) is the render bottleneck.** It loops in Python over 4,000 replicates × 10 classes, doing several boolean-mask reductions over 10,000-element arrays each time — roughly 40,000 iterations. Precomputing the per-class masks once and expressing recall/precision as matrix products against the `(B, n)` weight block would cut most of the ~110 s render. Correctness is not in question; this is purely cost.

**5. W7 has a latent degenerate case with no label.** `const b = Math.max(n - k, 1e-9)` guards against a class with perfect precision or recall. No current class triggers it (verified — closest is class 1 at 1121/1135), and I checked the numerics degrade gracefully rather than producing `NaN`: `betaInv(0.025, 1000, 1e-9)` returns `1`, `betaPdf ≈ 3.7e-7`. But the result would be an essentially empty density plot with no explanation. W4 already handles its degenerate case well — it draws the spike and labels it "posterior is a point mass" — and W7 could borrow that treatment.

**6. `rng_for` (`bootstrap.py`) derives a spawn key via `int.from_bytes(name.encode(), "little") % 2**63`.** Fine in practice for the short labels used, but a truncated `blake2b` digest would make the absence of collisions obvious rather than merely likely.

---

### Strengths

- **The verification is real, not decorative.** Ground truth was checked against the live data sources *before* the prose was written, and the build asserts it continuously. Exact enumerations are used wherever they're cheap — all 126 Efron patterns at n=5, the Poisson-conditioning/multinomial identity at n=3,4,5 to machine precision, the 35 equiprobable Dirichlet-multinomial patterns at n=4.

- **The hand-rolled JS special functions are correct.** I validated them against scipy in this review: `betaInv` matches to 8 decimal places on `(1,58)`, `(10,49)`, `(9687,313)`, `(1,49)`, `(190,10)`; `lgamma(0.5)` is exact to 10 places; `binomInterval(59, 1/59)` reproduces the `[0, 0.0508]` quoted in the post. Reimplementing these was the right call given the constraint, and it was done properly.

- **The weighted-likelihood-bootstrap scaling is right.** Verified: uniform weights reproduce unweighted training *exactly* (`4.5` vs `4.5`), and the weighted objective equals `sum_i w_i l_i` to float precision. That's the load-bearing claim of the MNIST section.

- **A genuine numerical bug was found and fixed rather than papered over.** `rng.gamma` underflows to exactly zero for small α, making the whole row normalise to `nan` — and α < 1 is precisely the edge-seeking regime the post's argument depends on. The log-space boost identity in `gamma_weights` is the correct fix and is documented at the call site.

- **`layout_qa.py` is a well-built gate**, including the two exclusions (`axison == False`, out-of-view tick labels) without which it would drown in false positives. It also handles the backend difference where `FigureCanvasBase` has no `get_renderer`.

- **Repo conventions are followed.** Frontmatter matches `CLAUDE.md`, `cover.png` is in house style, the dedicated venv is registered as a pinned kernel (verified: `kernel.json` `argv[0]` points at `.venv-bayesian-bootstrap/bin/python`, not the system interpreter), the theme is post-scoped so the rest of the blog is untouched, `docs/` is committed, and `data/` plus the venv are correctly gitignored.

- **Deviations are documented rather than hidden.** The README records the n=437-vs-438 discrepancy, the ~110 s render with the measured evidence for *why* `freeze` doesn't apply to single-file renders, and both upstream Quarto bugs — including the multi-statement `{ojs}` cell id mismatch, which is worth knowing beyond this repo.

---

### Verdict

**0 Critical, 2 Important, 4 Suggestions.** Both Important items are dead-code removals with no behavioural effect. Recommend merging and cleaning them up.

project-delphi:
### Both Important issues fixed in eb13a6a

Neither blocked merge, but both were dead code in a post whose premise is that the source is auditable, so they're gone rather than deferred.

1. **`index.qmd` dead code** — removed `n_metrics` and the `... if False else None` line entirely.
2. **`correct_rle`** — dropped the field and the `_rle` encoder. Every widget metric is a function of the per-class counts, so the 10,000 individual outcomes were bytes nothing read. `mnist.json` 2.2 KB → 0.8 KB; widget-data total 21.6 KB → 20.2 KB. (`NDArray` is still used by three other helpers, so that import stays.)

**Re-verified after the change:** render clean with all assertions passing · layout QA **10 figures, 0 collisions** · all 7 widgets OK in Chrome · W7 exercised specifically since its data source changed, reading correctly from the slimmed file (class 5: k=859 / n=892, point estimate 0.9630) · `correct_rle` confirmed absent from the rendered output.

The four Suggestions (`data.py` month heuristic, `metric_boot` vectorisation, W7's latent degenerate-case label, `rng_for` hashing) are left as-is — none affects correctness.

PR #119, opened 2026-07-27
Codify the branch/preview/data-context workflow and enforce no-commit-to-main

project-delphi:
## Summary

Three standing expectations existed only as habit. This writes them into `CLAUDE.md` and, for the one that actually has teeth, enforces it in the harness.

- **Never commit to `main`.** Documented as branch → commit → push → PR → review → merge, and backed by a `PreToolUse` hook on `Bash` (`.claude/hooks/block-main-commit.sh`) that denies any command reaching `git commit` while HEAD is on `main`/`master`. Previously nothing caught a stray commit.
- **Hand back a localhost preview link** when a unit of work is complete, and **kill Quarto preview servers once it merges** so stale previews don't hold ports.
- **Situate the data** in any post that touches a dataset: provenance, who collected it and why, the objective, downstream impact and cost of being wrong, and what property of the data makes the post's method the right tool. Synthetic data isn't exempt.

Also corrects a line in `## What this is` that said `docs/` is "committed directly to `main`" — now contradicted by the branch rule. It reads as served from `main`, rendered locally and committed in the same PR as the source.

### Design notes

- The hook matches **compound** commands (`cd docs && git commit …`), not just ones starting with `git`. An `if: "Bash(git *)"` filter was tried and dropped because it prefix-matches and would have let those through.
- Commands that **switch branch first** (`git switch -c rk/foo && git commit …`) are deliberately allowed — that is the correct move, so blocking it would be self-defeating.

## Test plan

- [x] Pipe-tested the script against blocked cases: `git commit -m x`, `git add -A && git commit -m x`, `git commit --amend --no-edit`, `git -C /dir commit`
- [x] Pipe-tested allowed cases: `git switch -c rk/foo && git commit -m y`, `git checkout -b rk/bar; git commit -m y`, `git status`, `gh pr merge 12`, `quarto render …`
- [x] Verified live on `main`: `git commit --allow-empty` was denied with the policy message, `git log -1` confirmed HEAD unchanged and no commit created
- [x] Verified the hook does *not* fire on a feature branch — the commit in this PR went through normally
- [x] `jq -e` validates the `settings.json` hook structure
- [ ] Not tested: hook loading in a fresh session (`.claude/` did not exist as a settings dir when this session started, so the config watcher may need `/hooks` or a restart to pick it up on first clone)

### DIFF
--- a/.claude/hooks/block-main-commit.sh
+++ b/.claude/hooks/block-main-commit.sh
@@ -4,17 +4,30 @@
 # Policy lives in CLAUDE.md: feature branch -> PR -> review -> merge.
 set -uo pipefail
 
-cmd=$(jq -r '.tool_input.command // ""')
-
-# Only intervene on commands that actually reach `git commit`, including
-# compound forms like `git add -A && git commit -m "..."` and `git -C dir commit`.
-if ! printf '%s' "$cmd" | grep -Eq '(^|[^[:alnum:]_./-])git([[:space:]]+[^;&|[:space:]]+)*[[:space:]]+commit([^[:alnum:]_-]|$)'; then
+# Fail open rather than breaking every Bash call, but say so — a policy hook
+# that silently stops enforcing is worse than one that is obviously off.
+if ! command -v jq >/dev/null 2>&1; then
+  echo "block-main-commit: jq not found; branch policy NOT enforced" >&2
   exit 0
 fi
 
-# A command that switches branch before it commits is exactly the right thing to
-# do, so let it through (e.g. `git switch -c rk/foo && git commit -m "..."`).
-if printf '%s' "${cmd%%commit*}" | grep -Eq 'git[[:space:]]+(switch|checkout|worktree)\b'; then
+cmd=$(jq -r '.tool_input.command // ""')
+
+# Only intervene when `commit` is the actual git subcommand — i.e. the first
+# token after `git` and any global options (`-C <path>`, `-c <cfg>`, `--no-pager`).
+# Anchoring this way still catches compound forms (`git add -A && git commit …`)
+# without snagging commands that merely mention the word (`git help commit`).
+commit_re='(^|[^[:alnum:]_./-])git([[:space:]]+(-[cC][[:space:]]+[^;&|[:space:]]+|--[^;&|[:space:]]+|-[^-;&|[:space:]]+))*[[:space:]]+commit([^[:alnum:]_-]|$)'
+if ! printf '%s' "$cmd" | grep -Eq "$commit_re"; then
+  exit 0
+fi
+
+# A command that *creates* a branch before committing is exactly the right thing
+# to do, so let it through (e.g. `git switch -c rk/foo && git commit -m "..."`).
+# Deliberately narrow: a bare `git switch main && git commit` must still be
+# blocked, so plain switches/checkouts do not count as an escape hatch.
+create_re='git[[:space:]]+(switch[[:space:]]+-[cC]|checkout[[:space:]]+-[bB]|worktree[[:space:]]+add)([^[:alnum:]_-]|$)'
+if printf '%s' "${cmd%%commit*}" | grep -Eq "$create_re"; then
   exit 0
 fi
 
--- a/.claude/hooks/block-main-commit.sh
+++ b/.claude/hooks/block-main-commit.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+# Blocks `git commit` while HEAD is on main/master.
+# Wired up as a PreToolUse hook on Bash in .claude/settings.json.
+# Policy lives in CLAUDE.md: feature branch -> PR -> review -> merge.
+set -uo pipefail
+
+cmd=$(jq -r '.tool_input.command // ""')
+
+# Only intervene on commands that actually reach `git commit`, including
+# compound forms like `git add -A && git commit -m "..."` and `git -C dir commit`.
+if ! printf '%s' "$cmd" | grep -Eq '(^|[^[:alnum:]_./-])git([[:space:]]+[^;&|[:space:]]+)*[[:space:]]+commit([^[:alnum:]_-]|$)'; then
+  exit 0
+fi
+
+# A command that switches branch before it commits is exactly the right thing to
+# do, so let it through (e.g. `git switch -c rk/foo && git commit -m "..."`).
+if printf '%s' "${cmd%%commit*}" | grep -Eq 'git[[:space:]]+(switch|checkout|worktree)\b'; then
+  exit 0
+fi
+
+branch=$(git rev-parse --abbrev-ref HEAD 2>/dev/null) || exit 0
+case "$branch" in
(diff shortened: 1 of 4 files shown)


project-delphi (commented):
## Review

### Overview

Writes three standing workflow expectations into `CLAUDE.md` (branch→PR→merge, hand back a preview link, situate the data in posts) and backs the first with a `PreToolUse` hook on `Bash` that denies `git commit` while HEAD is on `main`/`master`. Also corrects the now-contradictory "`docs/` is committed directly to `main`" line. Docs and enforcement land together, which is the right shape for this change.

Findings below were verified by running the script in a throwaway repo checked out on `main`, so the branch check actually engages.

---

### Critical Issues

**1. The escape hatch lets you switch *onto* `main` and then commit — bypassing the policy this PR exists to enforce.** (`.claude/hooks/block-main-commit.sh:16`)

The second guard exits 0 whenever the text before `commit` contains `git switch|checkout|worktree`. It never checks *what branch is being switched to*, so a command that moves onto `main` and commits there is allowed:

```
git switch main && git commit -m x         -> allowed   # should be blocked
git checkout main && git commit -m x       -> allowed   # should be blocked
git worktree list && git commit -m x       -> allowed   # should be blocked
```

`git switch main && git commit -m "..."` is a realistic thing to run, and it lands a commit directly on `main` with no prompt. The guard's stated intent is "a command that switches branch *before* it commits is the right thing to do" — but that reasoning only holds when the switch **creates a feature branch**. Narrowing the pattern to branch-*creation* forms (`switch -c/-C`, `checkout -b/-B`, `worktree add`) closes the hole and still permits the intended `git switch -c rk/foo && git commit …`.

The cost of narrowing is that `git switch <existing-feature-branch> && git commit` gets blocked while on `main`. That direction fails safe and the deny message says exactly what to do, so it is the right trade.

---

### Important Issues

**2. Silent fail-open when `jq` is absent.** (`.claude/hooks/block-main-commit.sh:7`)

`set -uo pipefail` omits `-e`, so if `jq` is missing, `cmd` is empty, the grep does not match, and the hook exits 0 — enforcement is silently off with no signal that the policy is no longer running. Fail-open is defensible for a hook (better than bricking every Bash call), but it should be *loud*. Consider a `command -v jq >/dev/null || { echo "block-main-commit: jq not found; branch policy NOT enforced" >&2; exit 0; }` so the degradation is visible.

---

### Suggestions

**3. False positives on read-only `git` commands.** The token-walk between `git` and `commit` is permissive enough to match subcommands that merely mention the word:

```
git help commit                            -> BLOCKED
git config --global commit.gpgsign false   -> BLOCKED
git log --grep commit                      -> BLOCKED
```

All harmless and fail-safe, but annoying. Requiring `commit` to be the actual subcommand — i.e. the first token after `git` and its global options (`-C <path>`, `-c <cfg>`, `--no-pager`, …) — removes all three while still catching `git -C dir commit` and `git --no-pager commit`.

**4. `git -C /other/repo commit` checks the wrong repository.** The branch is read via `git rev-parse` in the hook's cwd, not the `-C` target. Committing to a different repo while this one sits on `main` is wrongly blocked. Fails safe; a niche case, probably not worth the parsing.

**5. Redundant `pkill` in the CLAUDE.md teardown snippet.** `pkill -f "quarto preview"` is a strict subset of `pkill -f "quarto.*preview"` — the first is dead weight, and running both means the second matches nothing on the second pass. One pattern is enough.

**6. No automated test for the script.** It is 33 lines of shell that gates every commit; a handful of `printf … | hook.sh` assertions in a `.sh` next to it would lock in the allow/block matrix and catch regressions like #1. The repo has no test harness today, so this is a judgment call rather than a convention violation.

---

### Strengths

- **No injection surface.** The tool command is piped to `grep` via `printf '%s'` on stdin and never `eval`'d, interpolated into a shell string, or re-executed — the obvious way to write this hook is also the dangerous way, and it was avoided.
- **Correct deny contract**, and verified live rather than assumed: `permissionDecision`/`hookEventName` shape is right, the block was demonstrated on `main` with `git log -1` confirming HEAD unchanged.
- **Sensible fail-open on non-repo directories** (`git rev-parse … || exit 0`), so the hook does not break Bash calls outside a git checkout.
- **Compound commands are handled** (`cd docs && git commit …`), and the PR description explains why the cheaper `if: "Bash(git *)"` filter was tried and rejected — the reasoning is correct, that filter prefix-matches and would have leaked those.
- **Executable bit set** and `$CLAUDE_PROJECT_DIR` quoted in `settings.json`.
- The stale "committed directly to `main`" sentence was caught rather than left to contradict the new rule.

project-delphi (commented):
## Re-review after 481952b

Re-ran against the updated diff. **The Critical issue is resolved and no new Critical issues were introduced.**

### Critical Issues

None.

**1. Switch-to-main bypass — FIXED.** The escape hatch now exempts only branch *creation* (`switch -c/-C`, `checkout -b/-B`, `worktree add`). Verified the three previously-leaking commands are blocked, and the sanctioned path still passes:

```
git switch main && git commit -m x          -> BLOCKED   (was: allowed)
git checkout main && git commit -m x        -> BLOCKED   (was: allowed)
git worktree list && git commit -m x        -> BLOCKED   (was: allowed)
git switch -c rk/foo && git commit -m x     -> allowed   (unchanged, intended)
```

### Important Issues

None outstanding.

**2. jq fail-open — FIXED.** Missing `jq` now writes `branch policy NOT enforced` to stderr before exiting 0, so the degradation is visible instead of silent.

### Suggestions

**3. Read-only false positives — FIXED.** Anchoring `commit` as the real subcommand (after `git` and its global options) clears all three, while `git -C dir commit` and `git --no-pager commit` are still caught.

**6. Missing test — FIXED.** `.claude/hooks/test-block-main-commit.sh` asserts 22 cases across three regimes (on `main`, branch-creation forms, and on a feature branch), builds its own throwaway repo, and cleans up via `trap`. Commands are only ever fed to the hook as strings, never executed, so the suite has no side effects. Passes.

**7. Residual, new and minor: `git worktree add <path> <existing-branch> && git commit` is still exempt.** (`block-main-commit.sh:29`) The exemption matches `worktree add` without checking for `-b`, so adding a worktree on an existing branch and then committing — in the original cwd, still on `main` — slips through. Requires a fairly contrived command to hit, since the useful form is `cd`-ing into the new worktree first. Left as-is rather than adding more regex; noting it so it is a known edge rather than a surprise.

**4. Unchanged:** `git -C /other/repo commit` still reads the branch from the hook's cwd. Fails safe. Same reasoning as before — not worth the parsing.

**5. Redundant `pkill` — FIXED** in the CLAUDE.md teardown snippet.

### Notes

- `CLAUDE.md` was updated to describe the narrower rule accurately, including that switching to an *existing* branch and committing is blocked while on `main` — deliberate, fail-safe, and now stated rather than left for someone to discover.
- `shellcheck` is not installed in this environment, so the scripts were not statically linted; behavior is covered by the test suite instead.

**Verdict: no Critical issues. Clear to merge.**

PR #120, opened 2026-07-27
docs(claude): require a spine and stitched seams in every post

project-delphi:
## Summary

- Adds a **post structure** convention to `CLAUDE.md`: every post is one argument with a named core idea, claim-shaped headings (`## Efron's bootstrap is a weighted bootstrap in disguise`, not `## Background`), sections that each earn the next, and a conclusion that returns to the opening claim rather than summarising sections.
- The seam rule is the substantive part: a section's first sentence links back to the previous result or the core idea, its last names the unresolved thing the next section answers. Same for code blocks and figures. The test is falsifiable — if an opening sentence reads identically with the previous section deleted, the seam isn't stitched.
- Caveats go inline where the objection occurs, not into a "Limitations" bin at the end. This is what `posts/bayesian-bootstrap` already does; the convention now describes the house style instead of leaving it implicit.
- Compresses three pre-existing dense spots (branch-policy hook, preview handoff, venv/kernel setup) so the file stays skinny. `CLAUDE.md` loads into context every session, so length is a real cost.

**Net effect: 81 → 92 lines, 1354 → 1398 words.** The new section costs +44 words because the compression paid for it. Mid-work the file had ballooned to 1868 words before being trimmed back.

## Compression — nothing load-bearing dropped

This was a rewrite for density, not a deletion pass. All five venv failure modes survive, now as a three-line bash block plus the error string each command prevents:

- `ModuleNotFoundError: No module named 'yaml'` — why `ipykernel` alone isn't enough
- `ERROR: Jupyter kernel '<name>' not found` — why `QUARTO_PYTHON` is needed
- never system Python; `blog-base` for display-only posts; per-post `.venv-<slug>` + named kernel

Cut from the hook paragraph: the worked allow/block matrix. `.claude/hooks/test-block-main-commit.sh` asserts it, so the doc needn't restate it. Both script paths, the settings.json registration, the "switching to main is not an escape hatch" rule and both override routes are kept.

## Test plan

- [x] `wc -w/-l CLAUDE.md` → 1398 words / 92 lines, vs 1354 / 81 at the previous `main`
- [x] Grepped all 17 load-bearing strings still present: `block-main-commit.sh`, `test-block-main-commit.sh`, `QUARTO_PYTHON`, `nbclient`, `nbformat`, `pyyaml`, `blog-base`, `freeze: auto`, `#4A3AA7`, `output-dir: docs`, `title-block-banner`, `ipykernel install`, `eval: false`, `pkill`, `pgrep`, `langgraph-vs-llamaindex`, `ship-pr`
- [x] `.claude/hooks/test-block-main-commit.sh` → **PASS** (hook untouched; confirms the docs still match the script's actual behaviour)
- [x] Full `git diff` read end to end

No `docs/` rebuild accompanies this commit: `CLAUDE.md` is not rendered content, so there is no published page to drift and no preview URL for this change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,16 +18,14 @@ A personal ML/data blog ("Synthetic Musings") built with [Quarto](https://quarto
 
 **Never commit to `main`.** Every change — new post, edit, fix, even a one-line typo — goes through: feature branch → commit (source *and* the re-rendered `docs/` output together) → push → open a PR with a real description of what changed and why → PR review → merge. `main` only ever advances via a merged PR. The `ship-pr` skill automates this loop.
 
-This is enforced, not just documented: `.claude/settings.json` registers a `PreToolUse` hook on `Bash` that runs `.claude/hooks/block-main-commit.sh`, which denies any command reaching `git commit` while HEAD is on `main`/`master`. A command that *creates* a branch first passes (`git switch -c rk/foo && git commit …`); a bare `git switch main && git commit` does not — switching onto `main` is not an escape hatch. Switching to an *existing* branch and committing is also blocked while on `main`, which is deliberate: it fails safe, and the deny message says what to do. Run `.claude/hooks/test-block-main-commit.sh` after touching the hook — it asserts the full allow/block matrix. To override deliberately, commit from your own terminal, or disable the hook via `/hooks`.
+This is enforced: `.claude/settings.json` registers a `PreToolUse` hook on `Bash` running `.claude/hooks/block-main-commit.sh`, which denies any command reaching `git commit` while HEAD is on `main`/`master`. Creating a branch first passes (`git switch -c rk/foo && git commit …`); switching onto `main` is not an escape hatch. Run `.claude/hooks/test-block-main-commit.sh` after touching the hook — it asserts the full allow/block matrix. To override deliberately, commit from your own terminal or disable the hook via `/hooks`.
 
-**Hand back a localhost preview link whenever a unit of work is complete.** Don't just say "done" — give a clickable URL the change can be eyeballed at, e.g.:
+**Hand back a localhost preview link whenever a unit of work is complete** — a clickable URL, not just "done":
 
-- `quarto preview posts/<slug>/index.qmd` and report the URL it prints (typically `http://localhost:<port>/`), or
+- `quarto preview posts/<slug>/index.qmd`, reporting the URL it prints (typically `http://localhost:<port>/`) — prefer this single-post form, since a whole-project preview indexes and executes every post; or
 - serve the already-built output: `python -m http.server 8000 --directory docs` → `http://localhost:8000/posts/<slug>/index.html`.
 
-Prefer the single-post form; a whole-project preview indexes and executes every post.
-
-**Kill every running Quarto preview server once the change is shipped and merged.** After the PR merges, tear the servers down so stale previews don't linger on their ports:
+**Kill every running preview server once the change is merged**, so stale previews don't linger on their ports:
 
 ```bash
 pkill -f "quarto.*preview"        # quarto preview servers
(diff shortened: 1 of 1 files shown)


project-delphi (commented):
## Review — `docs(claude): require a spine and stitched seams in every post`

Single-file docs change (+21/−10 in `CLAUDE.md`). Two things happening at once: a new post-structure convention, and a compression pass over three pre-existing sections to pay for it. Reviewed both the new prose and — more importantly for a compression pass — whether anything load-bearing was silently dropped.

**Verdict: no Critical Issues. Merging.**

### Strengths

- **The convention is falsifiable, not vibes.** "If an opening sentence reads identically with the previous section deleted, the seam isn't stitched" is a test that can actually be applied to a draft. Same for "if two sections could swap without damage, merge or cut one." Most style guidance in this genre is unactionable; this isn't.
- **It describes the house style rather than inventing one.** `posts/bayesian-bootstrap` already uses claim-shaped headings (`## Efron's bootstrap is a weighted bootstrap in disguise`) and inline `## Caveat:` sections; `posts/tribes-and-calibration` opens on the core idea and closes by returning to it. Codifying what already works is much lower-risk than imposing a new structure on 60+ existing posts.
- **The compression fixed a latent contradiction.** The old `blog-base` line instructed `python -m ipykernel install --user --name blog-base` — a bare `python`, directly violating the **"Never install into or execute with the system Python"** rule stated one paragraph above it. The replacement ("register `blog-base` the same way") no longer contradicts the surrounding rule. That was a real bug in the docs, fixed incidentally.
- **Verification claims in the PR body check out.** 92 lines / 1398 words vs 81 / 1354 at `main`, all load-bearing strings present, hook test passes.

### Suggestions (non-blocking)

1. **`ipykernel` is missing from the explicit install line** (`CLAUDE.md:63`). The block installs `<post deps> jupyter nbclient nbformat pyyaml`, then line 64 invokes `-m ipykernel install`. This works today only transitively — `pip show jupyter` gives `Requires: ipykernel, ipywidgets, jupyter-console, jupyterlab, nbconvert, notebook`. The risk is that `jupyter` is a heavyweight metapackage (drags in jupyterlab + notebook), so a future reader optimizing the install is quite likely to drop it — and silently break kernel registration. Naming `ipykernel` explicitly makes the block self-evidently correct and costs one word. The prose even says "`ipykernel` alone is not enough", which reads as though it's in the list.

2. **The kernel-misregistration diagnostic was dropped.** The old text ended with: verify via `jupyter kernelspec list` and check the kernel's `kernel.json` `argv` path if a render picks up the wrong environment. That was the only debugging aid for "the kernel exists but points at the wrong interpreter" — a failure that presents as confusing `ImportError`s rather than a clean error string, so it's the hardest of the documented modes to diagnose unaided. The other four modes survive because each has a greppable error message; this one didn't. Worth one clause if it recurs.

3. **Nit** — the first preview bullet (`CLAUDE.md:25`) now carries a long mid-clause parenthetical and terminates in `; or`, which reads awkwardly at that length. Cosmetic.

### Risk / correctness

- All five documented venv failure modes survive the rewrite, each still paired with its literal error string (`ModuleNotFoundError: No module named 'yaml'`, `ERROR: Jupyter kernel '<name>' not found. Known kernels: python3`).
- Hook paragraph: both script paths, the `settings.json` registration, the "switching onto `main` is not an escape hatch" rule and both override routes are retained. Dropping the worked allow/block matrix is safe — `.claude/hooks/test-block-main-commit.sh` is the executable spec and still passes.
- No secrets, no credentials, no executable code paths touched. `CLAUDE.md` is excluded from the Quarto render, so there is correctly no accompanying `docs/` output and no risk of source/site drift.
- Only real cost of being wrong here is context-window budget every session — and that moved the right way (+44 words net for a whole new convention).

### Test coverage

Appropriate for the change. `.claude/hooks/test-block-main-commit.sh` passes (hook itself untouched; the run confirms the *compressed description* still matches the script's actual allow/block behaviour, which is the thing that could have rotted).

PR #121, opened 2026-07-27
build: adopt uv for tooling, lock every post's dependencies

project-delphi:
## Summary

Switches the base tooling from `python -m venv` + `pip` to [uv](https://github.com/astral-sh/uv), and captures a pinned lockfile for every post that has a dedicated venv.

**Migration is deliberately lazy.** Only the base `.venv` is rebuilt here. The nine post venvs are left untouched and keep working; each migrates the next time its post needs re-execution.

### Why this is safe

Jupyter kernel specs store an **absolute interpreter path**:

```
~/Library/Jupyter/kernels/tda-blog/kernel.json
  argv[0] = /Users/ravikalia/Code/github.com/ml-blog/.venv-tda/bin/python
```

`uv venv .venv-tda` puts the interpreter at exactly that path, so recreating a venv in place needs no kernel re-registration, no frontmatter change, and no `docs/` rebuild. The `jupyter: <kernel-name>` → `kernel.json` → interpreter chain is uv-agnostic. Verified end to end below.

### Why lock the versions

`_freeze/` is gitignored (0 tracked files) while `docs/` is committed. A re-render on drifted dependencies would silently change published output with **no record of what produced the old one**. Nine posts now carry a `requirements.txt` capturing the full transitive set from the working venv, with restore instructions in the header.

`posts/bayesian-bootstrap/requirements.txt` is deliberately **not** overwritten — it is a curated, commented, direct-dependency list, strictly better than a freeze dump. All 15 of its pins were verified to match the installed venv exactly.

### Motivation

Ten venvs currently occupy **5.3 GB**, all on Python 3.13.2, all duplicating numpy/scipy/matplotlib and in three cases torch. uv hardlinks from a shared global cache, so the duplication collapses as venvs are recreated.

## Two pre-existing bugs fixed

1. **`.gitignore` only matched `.venv` exactly.** Every `.venv-*` was untracked purely because Python 3.11+ writes a `.gitignore` containing `*` *inside* each venv. Now `.venv*`, so 5.3 GB cannot be swept in if a venv ever loses that internal file.
2. **`posts/bayesian-bootstrap/requirements.txt` claimed `make venv` "also registers the Jupyter kernel".** It did not — the target only ran `python3 -m venv`. There is now a `make kernel` that actually does it. (The old target's `source .venv/bin/activate` was also a no-op, since each Make recipe line runs in its own shell.)

Also folds in the two suggestions from the #120 review: `ipykernel` is now named explicitly in the install line rather than arriving transitively via the `jupyter` metapackage, and the kernel-misregistration diagnostic (`jupyter kernelspec list` + `argv[0]`) is restored.

## Test plan

- [x] `brew install uv` → 0.11.32. Probed in a scratch dir first: `uv venv` **does** write a self-ignoring `.gitignore`, and a venv without `--seed` has **no** `pip` — both assumptions this PR's docs rest on, confirmed before writing them
- [x] `make venv && make install && make kernel` from scratch (deleted `.venv` first) — succeeds; `blog-base` re-resolves to `<repo>/.venv/bin/python`
- [x] End-to-end render through the uv-built venv: `QUARTO_PYTHON="$(pwd)/.venv/bin/python" quarto render posts/langgraph-vs-llamaindex/index.qmd` → `Output created`. Display-only post, so it proves Quarto gets a working kernel without executing anything
- [x] That render's `docs/` churn was **verified cosmetic and reverted** — regenerated cell `id` attributes and one mtime sort key; byte-identical after normalizing those. No source change to that post, so no `docs/` update belongs in this PR
- [x] All 12 registered kernels still resolve; the 9 untouched post venvs are unaffected
- [x] All 15 curated `bayesian-bootstrap` pins match the installed venv
- [x] `git status` shows no `.venv*` path; new-file credential scan is clean (only match was the package `asttokens`)
- [x] `.claude/hooks/test-block-main-commit.sh` → PASS
- [x] `CLAUDE.md` 92 → 95 lines, 1398 → 1468 words

## Known follow-up (not in this PR)

**`posts/langchain` pins `jupyter: python3`, and that kernel resolves to `/Users/ravikalia/Code/local-playground/.venv/bin/python` — a venv in an unrelated repository.** That post is not reproducible on any other machine and its committed `docs/` cannot be regenerated from this repo alone. Fixing it needs its own venv built and the post re-rendered, which is real `docs/` churn and deserves a separate PR.


project-delphi (commented):
## Review — `build: adopt uv for tooling, lock every post's dependencies`

13 files, +986/−12 — of which ~960 lines are generated lockfiles. The reviewable surface is `Makefile`, `.gitignore`, `CLAUDE.md`, and the *shape* of the generated files.

**Verdict: no Critical Issues. Merging.** One Important finding below that will quietly undermine the convention this PR introduces, plus three minor ones.

### Strengths

- **The central safety argument is correct and was actually verified, not assumed.** Kernel specs store `argv[0]` as an absolute interpreter path, and a same-named `uv venv` satisfies it unchanged — which is what makes lazy migration possible at all. All 12 registered kernels still resolve.
- **uv's behaviour was probed in a scratch directory before being written into the docs** — that `uv venv` self-ignores and that a non-seeded venv has no `pip`. Documenting unverified tool behaviour is the usual way this kind of PR goes wrong.
- **The generated lockfiles avoid every standard `pip freeze` trap**: no `@ file://` URLs, no `-e` editable installs, no local build tags (`2.13.0+cpu`). Every non-comment line is a plain `name==version`, so they will actually restore on another machine.
- **`posts/bayesian-bootstrap/requirements.txt` was correctly left curated** rather than clobbered with a 200-line freeze dump, and its 15 pins were checked against the installed venv.
- **The cosmetic `docs/` churn was reverted rather than committed.** Regenerated cell ids and an mtime key are noise; committing them would have made a tooling PR look like a content PR.

### Important

**`CLAUDE.md` now asserts a lockfile convention it never tells anyone how to follow.** The new sentence reads "Each post's exact versions are locked in `posts/<slug>/requirements.txt`", but the setup block immediately above it — the thing someone actually copies when adding a post — ends at `quarto render`. `grep -c "pip freeze" CLAUDE.md` → **0**.

So the very next code-heavy post created by following these instructions gets a venv, a kernel, and no lockfile, and the convention starts decaying from post #13 onward. The reproducibility argument in the PR description is the whole justification for the 960 added lines; it needs one more line in the block to survive contact with the next post:

```bash
uv pip freeze --python .venv-<slug>/bin/python > posts/<slug>/requirements.txt
```

### Suggestions

1. **The "each post" claim over-generalizes.** Two of the twelve code posts have no lockfile, both legitimately: `posts/langgraph-vs-llamaindex` (shares `blog-base`) and `posts/langchain` (the cross-repo `python3` kernel already flagged as follow-up). Scoping the sentence to *posts with a dedicated venv* would keep it true.

2. **`make install` has no `venv` prerequisite.** On a fresh clone it fails with a raw uv error about a missing interpreter. Worth making it `install: venv` — and that is safe now, because `uv venv` on an existing venv is **non-destructive**: it declines and hints `--clear` (verified with a canary file; the existing environment survived untouched). Worth knowing independently, since it means `make venv` can't nuke a working environment.

3. **The two `.venv-tda` lockfiles are duplicates kept in sync by a header comment only.** Currently byte-identical (verified), but nothing enforces it. Acceptable given the post-independence architecture — just noting the drift path exists.

### Risk

- The headline disk win is **mostly unrealized today** — the base venv went 368M → 321M, against 5.3 GB total — because lazy migration was the chosen scope. The PR body hedges this correctly ("collapses as venvs are recreated"), so this is framing-accurate, not a defect. The reproducibility gain is the real deliverable here.
- Nine post venvs are untouched, `_freeze/` is intact, and no post was re-executed, so there is no path from this change to altered published output.
- `.venv*` cannot over-match anything tracked. Strictly safer than depending on the `.gitignore` Python writes inside each venv.

### Test coverage

Strong for a build change, and the end-to-end render is the right test: `posts/langgraph-vs-llamaindex` executes nothing but still requires a working kernel, so it exercises the full `QUARTO_PYTHON` → kernel → interpreter chain against the uv-built venv without risking content drift.

project-delphi:
## Follow-up: review findings applied in `d91e2e1`

**Important finding fixed.** The lockfile generation step is now part of the setup block in `CLAUDE.md`, so a new post actually produces one:

```bash
uv pip freeze --python .venv-<slug>/bin/python > posts/<slug>/requirements.txt
```

The surrounding sentence now also states *why* it isn't optional (`_freeze/` gitignored, `docs/` committed) and is scoped to posts with a dedicated venv, which drops the over-generalization noted in suggestion 1.

**Correction to suggestion 2 in the review above.** I wrote that `uv venv` on an existing venv "declines and hints `--clear`", and concluded `install: venv` was therefore safe. The first half is right — the environment is genuinely not destroyed, canary intact — but it declines by **exiting non-zero**, not by no-opping:

```
$ make install
hint: Use the `--clear` flag or set `UV_VENV_CLEAR=1` to replace the existing virtual environment
make: *** [venv] Error 2
```

So adding the prerequisite as suggested broke `make install` on every existing checkout. The correct form is `uv venv --allow-existing .venv`, which exits 0 and leaves the installed toolchain importable. Both are in `d91e2e1`.

Re-verified after the fix: `make install` against an existing venv succeeds, `make venv` twice in a row exits 0 both times, `make kernel` still registers `blog-base` at the right interpreter.

PR #122, opened 2026-07-27
fix(claude): catch docs servers started with cd instead of --directory

project-delphi:
## Summary

The documented preview teardown had a blind spot that let a stale server survive — and it bit for real this session.

`pkill -f "http.server .*docs"` only matches servers started as `python -m http.server --directory docs`. A server started the other common way:

```bash
cd docs && python -m http.server 8899
```

has **no `docs` anywhere in its command line**, so the pattern never matched it. It sat on port 8899 through several merges, and I initially misreported it as belonging to another project on exactly that basis.

The fix matches on **working directory** instead of argv, which catches both forms and — as a bonus — is narrower where it matters: it deliberately spares `http.server` processes rooted outside this repo, which a broader `pkill -f "http.server"` would have killed.

```bash
pgrep -f "http\.server" | while read -r pid; do
  lsof -a -p "$pid" -d cwd -Fn 2>/dev/null | grep -q "^n$PWD" && kill "$pid"
done
```

The confirmation step is extended to match, since `pgrep -fl "quarto preview"` alone never covered the static-server case.

## Test plan

Tested against the exact case that slipped through, with a real server started the `cd docs` way:

- [x] Confirmed the blind spot first: with the test server live, `pgrep -f "http.server .*docs"` returns **no match**
- [x] Dry run of the new logic correctly partitions the two live servers — `WOULD KILL` the one with `cwd=<repo>/docs`, `would skip` an unrelated one at `cwd=/private/tmp/bbsmoke`
- [x] Live run: repo-rooted server killed, unrelated server on 8900 untouched
- [x] Re-ran with the snippet **extracted verbatim from `CLAUDE.md`** (`sed`'d out of the file and executed) rather than retyped, so the documented text is what was actually tested
- [x] `.claude/hooks/test-block-main-commit.sh` → PASS
- [x] `CLAUDE.md` 96 → 101 lines, 1505 → 1574 words

Docs-only change; `CLAUDE.md` is not rendered content, so no `docs/` rebuild.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,10 +29,15 @@ This is enforced: `.claude/settings.json` registers a `PreToolUse` hook on `Bash
 
 ```bash
 pkill -f "quarto.*preview"        # quarto preview servers
-pkill -f "http.server .*docs"     # any static docs server started for previewing
+# Static docs servers. Matching argv misses `cd docs && python -m http.server`
+# (no "docs" in its command line), so match on cwd — which also spares servers
+# belonging to other projects. Run from the repo root.
+pgrep -f "http\.server" | while read -r pid; do
+  lsof -a -p "$pid" -d cwd -Fn 2>/dev/null | grep -q "^n$PWD" && kill "$pid"
+done
 ```
 
-Confirm nothing survives (`pgrep -fl "quarto preview"` should print nothing) and say so in the wrap-up.
+Confirm nothing survives — `pgrep -fl "quarto preview"` prints nothing, and `lsof -nP -iTCP -sTCP:LISTEN | grep -i python` lists no server rooted in this repo — and say so in the wrap-up.
 
 ## Writing conventions
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,8 @@ pkill -f "quarto.*preview"        # quarto preview servers
 # (no "docs" in its command line), so match on cwd — which also spares servers
 # belonging to other projects. Run from the repo root.
 pgrep -f "http\.server" | while read -r pid; do
-  lsof -a -p "$pid" -d cwd -Fn 2>/dev/null | grep -q "^n$PWD" && kill "$pid"
+  cwd=$(lsof -a -p "$pid" -d cwd -Fn 2>/dev/null | sed -n 's/^n//p')
+  case "$cwd" in "$PWD"|"$PWD"/*) kill "$pid";; esac
 done
 ```
 


project-delphi (commented):
## Review — `fix(claude): catch docs servers started with cd instead of --directory`

Docs-only, one file, +7/−2. The change is a shell snippet in `CLAUDE.md` that agents copy and run against live processes, so it was reviewed as executable code rather than prose.

**One Important issue found and fixed in `2fa0c02` before merge.** No Critical.

### Important (fixed)

**The `grep -q "^n$PWD"` match was an unanchored regex, and over-reached in two ways.** Both verified empirically, not theorized:

| cwd of an `http.server` | as shipped in `95f5211` |
|---|---|
| `…/ml-blog` | kill (correct) |
| `…/ml-blog/docs` | kill (correct) |
| `…/ml-blog-drafts/docs` | **kill — wrong** |
| `…/githubXcom/ml-blog/docs` | **kill — wrong** |
| `/private/tmp/bbsmoke` | spare (correct) |

1. **Prefix, not path boundary.** `$PWD` = `…/ml-blog`, so any sibling whose name merely *starts with* `ml-blog` matched. A preview server in a `ml-blog-drafts` checkout would have been killed by a teardown run from this repo.
2. **`$PWD` interpolated into a regex.** The `.` in `github.com` is a wildcard, so a literally different path matches.

That is the same class of error the PR set out to fix — killing or sparing the wrong server — reintroduced in a new form, and it directly contradicted the PR body's claim to "deliberately spare `http.server` processes rooted outside this repo." Replaced with a literal, boundary-anchored glob:

```bash
cwd=$(lsof -a -p "$pid" -d cwd -Fn 2>/dev/null | sed -n 's/^n//p')
case "$cwd" in "$PWD"|"$PWD"/*) kill "$pid";; esac
```

`case` globs are literal (no regex metacharacters), `"$PWD"` is quoted so its content isn't glob-expanded, and the `|"$PWD"/*` alternation anchors on the path separator. Re-verified against all five paths above — correct on every one.

### Strengths

- **The underlying diagnosis is right and was demonstrated, not asserted.** With a server started `cd docs && python -m http.server`, the old `pgrep -f "http.server .*docs"` returns no match. That is a genuine hole, and it had actually let a server survive several merges.
- **Matching on cwd is the correct fix**, and strictly better than the obvious alternative — a broad `pkill -f "http.server"` would have killed the unrelated `/private/tmp/bbsmoke` server.
- **The snippet was tested by extracting it from `CLAUDE.md` with `sed` and executing that**, rather than retyping it. For a file whose entire purpose is to be copied and run, testing the artifact instead of a paraphrase of it is the right discipline — and it is what caught that the documented text runs clean.

### Minor (not blocking)

- **`pkill -f "quarto.*preview"` on the line above is still repo-blind** and would kill Quarto previews belonging to any other project. Pre-existing and out of scope here, but the block is now inconsistent: the second command is carefully repo-scoped while the first is not.
- **The confirmation command doesn't filter by repo.** `lsof -nP -iTCP -sTCP:LISTEN | grep -i python` lists *every* Python listener, while the sentence describes it as showing "no server rooted in this repo" — so it needs eyeballing rather than being a clean pass/fail. Acceptable for a confirmation aid.
- `pgrep` exits non-zero when nothing matches, so the snippet would abort a `set -e` script. It is documented as an interactive copy-paste block, so this is fine as-is.

### Risk

Low and bounded. Worst case for the bug above was killing a static file server in a sibling checkout — no data loss, restartable with one command — and it is fixed. No repo content, build, or render path is touched; `CLAUDE.md` is excluded from the Quarto render, so there is correctly no `docs/` output in this PR.

PR #123, opened 2026-07-27
feat(posts): reproducibility-ladder post; install base toolchain from uv.lock

project-delphi:
Two related changes from one session. They're separable — say the word and I'll split them.

## Summary

**New post — "'It Works on My Machine': The Ladder of Numerical Dependencies"**

- Organizes the whole piece around a *ladder of dependencies* (chipset → OS → runtime → packages → application code). The load-bearing claim is that a fix at one rung doesn't fix the rungs around it, and most tools own one or two rungs while implying they own the ladder.
- That framing is what lets the devops stack and the research-computing stack be read as answers to the same problem under different resource constraints, rather than as rival toolkits. The resource asymmetry (no standard target, no platform team, a 5–10 year horizon) is the pivot between them.
- Time-sensitive claims were researched rather than recalled. Anthropic/Bun (Dec 2025), OpenAI/Astral (Mar 2026), OpenAI/Ona (Jun 2026) — the latter two still pending close and worded as "agreed to acquire".
- Prose only, no executable cells, so no dedicated venv or kernel is needed.

**Base toolchain now installs from the lock**

- `make install` was `uv pip install .` — a fresh resolve on every clone. The repo asserted a lockfile convention for per-post venvs while its own base environment had none.
- Now `uv sync`, installing from a committed root `uv.lock`.
- `uv sync` *prunes* anything not in the lock. That's the point, but it's the one behavior change that can bite someone who hand-installed into `.venv`, so CLAUDE.md now warns about it.
- `make lock` kept, rescoped to regenerating the lock without touching `.venv`.

## Test plan

- [x] `quarto render posts/numerical-computing-reproducibility/index.qmd` — clean; only this post plus the homepage listing rebuilt, no other post's output touched
- [x] New title propagates to page `<title>`, homepage listing, and `search.json`
- [x] 24 of 26 external links return HTTP 200; the two `openai.com` URLs 403 to `curl` (bot-blocking, not rot) and were confirmed as the official announcement pages during research
- [x] All eight languages covered with tool names; section spine checked by reading only each section's first and last sentence
- [x] `make install` idempotent across repeated runs; `ipykernel`, `ruff`, `black`, `mypy`, `pre-commit` all resolve afterward
- [x] Fresh-clone path simulated via `UV_PROJECT_ENVIRONMENT` into a throwaway venv — builds from the lock alone with matching versions
- [x] `blog-base` kernel spec still resolves (absolute path into `.venv` unchanged)
- [x] `uv.lock` scanned for embedded credentials — clean

## Notes for the reviewer

- Body is ~2,690 words against a ~1,800–2,500 target. Two trimming passes; going further meant cutting mandated content (eight languages, five deployment targets, three failure modes).
- `ml_blog` is not importable after the switch — it never was. There's no `ml_blog/` source dir; the project is a pure dependency container. Verified identical under the old install path.
- `uv sync` is used without `--locked`, so it auto-updates the lock on `pyproject.toml` drift. Right for local dev; `--locked` would be the CI-grade choice if you'd rather it fail loudly.


project-delphi (commented):
## Review — PR #123

**Overview.** Two changes: a new prose-only Quarto post organized around a "ladder of dependencies," and a build change making `make install` install from a committed root `uv.lock` via `uv sync` instead of re-resolving with `uv pip install .`. 10 files, but the reviewable surface is small — `uv.lock` (5,720 lines) and the rendered `docs/` output are generated artifacts, and committing `docs/` is required by this repo's convention.

### Critical Issues

None.

### Important Issues

**1. Section heading now contradicts its own body — `posts/numerical-computing-reproducibility/index.qmd:78`**

The heading reads *"Eight languages, eight different sets of rungs left unpinned"*, but the C/C++/Fortran entries were merged into one paragraph (line 88) that opens *"**C, C++, and Fortran** share the deepest problem"*. The body now presents three of the eight languages as sharing **one** set of rungs, directly contradicting the heading's "eight different sets."

Eight languages are still named, so the first half holds; it's the parallel claim that broke. Headings in this repo are supposed to state claims that survive scrutiny (CLAUDE.md: "headings state claims, not topics"). Suggest something like *"Every language pins a different part of the ladder"* — keeps the claim, drops the now-false count.

**2. `pyproject.toml` has no package configuration, so setuptools auto-discovers `posts/` as a package**

Not introduced here, but this PR changes how it manifests and makes it worth fixing. Verified in the synced venv:

```
$ .venv/bin/python -c "import posts; print(posts.__path__)"
IMPORTABLE: _NamespacePath(['/Users/.../ml-blog/posts', ...])
```

The old `uv pip install .` was non-editable, so setuptools **copied** the entire `posts/` tree into `site-packages` (the stale `ml_blog.egg-info/top_level.txt` still records `posts`). `uv sync` installs editable instead, so this is now a path hook rather than a copy — strictly an improvement, but the underlying misconfiguration remains and now silently puts the repo root on `sys.path`. A one-line `[tool.setuptools] py-modules = []` (or `packages = []`) would make the project the pure dependency container it's intended to be.

### Suggestions

**3. `install: venv` prerequisite is now redundant — `Makefile:6`**

`uv sync` creates and manages `.venv` itself, so `uv venv --allow-existing .venv` no longer does necessary work. Harmless today, but on a clone whose default interpreter doesn't satisfy `requires-python`, `uv venv` would create a venv that `uv sync` then has to replace. Dropping the prerequisite would make `install` a single source of truth.

**4. `requires-python = ">=3.8"` constrains the lock to EOL-compatible versions**

Python 3.8 has been end-of-life since October 2024. Because the lock is universal, this floor holds resolution back to versions still supporting 3.8 across all 244 packages. The venv actually in use is 3.13. Raising the floor would let the lock resolve to current releases — relevant now that the lock is authoritative rather than decorative.

**5. Help text drifts from the Makefile comment — `Makefile:27`**

`@echo "lock - refresh the root uv.lock..."` says "refresh," while the comment above the target says "regenerate ... without touching .venv." The comment is the more precise description; worth aligning so the distinction from `install` survives.

### Strengths

- The `uv sync` switch is the change that makes the existing lockfile convention actually load-bearing rather than aspirational — it closes a real gap where the repo mandated lockfiles for per-post venvs while its own base environment re-resolved on every clone.
- The pruning behavior of `uv sync` is the one genuinely surprising side effect, and it's documented at the point of use in CLAUDE.md rather than left to be discovered.
- Time-sensitive claims in the post are sourced to primary announcements, and the two acquisitions still pending regulatory close are worded as "agreed to acquire" rather than completed — an easy detail to get wrong.
- Commits are correctly split along their two concerns, with post source and rendered `docs/` landing together as CLAUDE.md requires.
- Further reading has 10 links across 4 topic groups; external links verified (24/26 return 200, the two `openai.com` 403s are bot-blocking, not rot).

### Coverage / risk notes

No test suite exists in this repo, so verification is necessarily manual; the PR's test plan covers the meaningful paths, including a fresh-environment build from the lock via `UV_PROJECT_ENVIRONMENT`. No security concerns — `uv.lock` carries only pypi.org URLs with SHA-256 hashes and no credentials or private index references, and adding hash-pinned resolution is a net supply-chain improvement.

project-delphi:
Fixed Important #1 in d770ed1 — heading is now "Every language leaves a different rung unpinned," which no longer asserts a count the merged C/C++/Fortran paragraph contradicts. Re-rendered; TOC and search index updated.

Important #2 (setuptools auto-discovering `posts/`) and Suggestions #3–#5 left as follow-ups — #2 is a `pyproject.toml` change outside this PR's scope, and the rest are non-blocking.

project-delphi:
Suggestions #3–#5 fixed in f86461d.

**#4 (`requires-python`) — I need to correct what I claimed in the review.** I said the 3.8 floor was "holding resolution back to versions still supporting 3.8." That was wrong: `ruff 0.16.0`, `black 26.5.1`, and `mypy 2.3.0` all resolved to current releases on 3.13 regardless. The actual cost was lock *forking* — 64 of 244 packages carried duplicate pinned versions to serve interpreters this repo never uses. Raising the floor to `>=3.11` collapses that:

| | before | after |
|---|---|---|
| packages | 244 | 132 |
| forked (>1 version) | 64 | 0 |
| lockfile lines | 5720 | 2333 |

Still worth doing — one version per package means fewer ways for two clones to diverge — but the benefit is determinism, not freshness.

**#3** — `install` no longer depends on `venv`; `uv sync` creates `.venv` itself. `make venv` kept for a bare venv.

**#5** — help text now distinguishes `lock` (leaves `.venv` alone) from `install`.

CLAUDE.md updated to match, including dropping `make venv &&` from the documented kernel-setup flow.

Verified: install idempotent; fresh environment builds from the lock with no pre-made venv; `make install && make kernel` still registers `blog-base`.

Important #2 (setuptools auto-discovering `posts/`) still open — separate concern.

project-delphi:
Important #2 fixed in f6b9f6c — `[tool.setuptools] packages = [] / py-modules = []` disables flat-layout auto-discovery.

**A measurement caveat worth recording**, since it nearly fooled me: `python -c` puts the cwd on `sys.path`, so `import posts` succeeds from the repo root whether or not the install exports it. Testing from the root is useless here. The signal that the install was actually at fault was the finder entry in the path list:

```
# before
_NamespacePath(['.../ml-blog/posts', '.../ml-blog/posts',
                '__editable__.ml_blog-0.1.0.finder.__path_hook__'])
# after (from /tmp)
ModuleNotFoundError: No module named 'posts'
```

Also confirmed the editable finder no longer references `posts` and dist-info `top_level.txt` is empty.

Verified in a fresh environment: `posts` not importable, ruff/ipykernel intact, `make kernel` still registers `blog-base`.

All review findings from #123 are now closed (1 fixed pre-merge, 2 Important, 3 Suggestions).

PR #127, opened 2026-07-29
fix(posts): retract the "no rush" migration advice in Skills Ate Commands

project-delphi:
## Summary

- The closing section of "Skills Ate Commands" told readers not to bother migrating `.claude/commands/` to skills. That answer was priced for a world where migration meant an afternoon of clerical work per command — which is no longer the trade.
- Strikes the old paragraph through (`~~…~~` → `<del>`) rather than deleting it, so the change of mind is visible rather than quietly rewritten.
- Adds the pushback from Scott Mountenay (LinkedIn linked) as a blockquote, verbatim except `create a or` → `create a [PR]`, bracketed to mark the correction.
- Follows with why he's right, a concrete bulk-migration prompt, and the one judgement call the reviewer has to make: which commands still need `disable-model-invocation` now that their descriptions sit in context and Claude can fire them unprompted.
- Re-stitches the "write anything *new* as a skill" close so it reads as continuation, not contrast.

## Test plan

- [x] `QUARTO_PYTHON=.venv-skills-vs-commands/bin/python quarto render posts/skills-vs-commands/index.qmd` — all 4 cells executed, output created
- [x] Rendered HTML contains `<del>` for the struck paragraph and the LinkedIn link
- [x] Eyeballed the section over `python -m http.server` on the built `docs/`
- [x] `docs/` output committed alongside the source change


project-delphi (commented):
## Review

**Overview** — Prose-only change to the closing section of `posts/skills-vs-commands/index.qmd`, plus the correspondingly re-rendered `docs/`. Retracts the "don't bother migrating" advice by striking it through rather than deleting it, quotes the pushback that prompted the change with attribution, then supplies the replacement advice, a bulk-migration prompt, and the one judgement call the migration leaves to a human. No code, no dependency, no config changes.

### Critical Issues

None.

### Important Issues

- **Stale search-index entries still serve the retracted advice, at a URL that 404s.** `docs/search.json` carries 9 entries under `posts/skills-vs-commands/skills-vs-commands.html#…` — an orphan from an older build; only `index.html` exists in that directory now. The one covering this section still contains the *pre-retraction* text verbatim. This is pre-existing on `main`, not introduced here, and a targeted `quarto render posts/<slug>/index.qmd` can't clear it (only a full `quarto render .` rebuilds `search.json` from scratch). Flagging it because it lands squarely on this PR's purpose: site search can still surface the un-retracted paragraph, and the result it links to is a dead page. Worth a separate full-render PR.
- **"That was the honest trade a year ago" sits awkwardly against the post's own date.** The post is dated `2026-07-24` — five days ago — so a reader who checks the date sees advice published last week described as the state of the world a year back. The sentence means "the era that advice was written for," but the byline date argues with it. `"That was the honest trade until recently"` (or naming what changed rather than when) removes the friction without losing the point.

### Suggestions

- Quarto strips markup when building the search index, so even the *correct* `index.html` entry opens with the struck paragraph as plain text. Unavoidable with strikethrough-in-place; a one-line "**Update:**" lead-in before the struck paragraph would at least survive into the indexed text.
- The `[PR]` in the blockquote is an editorial correction of `create a or` in the original message. The brackets signal that to a careful reader, but nothing states it. A trailing "(bracketed word mine)" would make the convention explicit.
- Substantive change to a published post with no frontmatter `date` bump. There's no repo convention either way — noting it in case you want the homepage listing to resurface the post.

### Strengths

- Striking through rather than deleting is the right call: the change of mind is the content here, and a silent rewrite would erase it.
- Seams are stitched per the repo's writing conventions — "Struck through, because…" links back to the deleted claim, and the final paragraph was re-opened with "And" so it reads as continuation rather than the contrast the original "But" set up against advice that no longer stands.
- Counterpoint is credited by name with a link, not absorbed as the author's own reasoning.
- `docs/` committed in the same commit as the source, per `CLAUDE.md`. The re-render is clean: the post's generated artifacts (`contrast-chart.png`, `contrast-report.html`, `sample.css`) came back byte-identical, so the diff is confined to the changed prose plus expected reading-time/timestamp churn.

PR #128, opened 2026-07-29
fix(posts): clear stale search entries and the date-anchored aside

project-delphi:
## Summary

Both follow-ups from the review on #127.

**1. Stale search-index entries (the Important one).** `docs/search.json` carried 9 entries under `posts/skills-vs-commands/skills-vs-commands.html` — the post's filename at its first commit (`04899b5`), never cleaned up since. Two consequences: site search could return the *pre-retraction* text that #127 just struck through, and the result linked to a page that doesn't exist in `docs/`.

Pruned rather than rebuilt. A full `quarto render .` is the "correct" fix, but only 31 of 71 posts have a `_freeze/` cache locally (`_freeze/` is gitignored), so it would re-execute ~40 posts needing R/PySpark/HuggingFace kernels — against a `CLAUDE.md` warning not to do that for a single-post change. The 9 orphans duplicate the live `index.html` entries section for section, so dropping them is exactly what a clean rebuild would produce. Verified afterwards that every remaining `href` resolves to a file that exists in `docs/`.

**2. The date-anchored aside.** "That was the honest trade a year ago" argued with the post's own `2026-07-24` byline. Re-anchored to the condition that changed — "for as long as the clerical work was yours to do" — which is the actual claim and doesn't rot.

## Test plan

- [x] `QUARTO_PYTHON=.venv-skills-vs-commands/bin/python quarto render posts/skills-vs-commands/index.qmd` — 4/4 cells executed, output created
- [x] Re-render preserves the pruned index: 508 entries, no orphans re-added
- [x] Every `href` in `search.json` resolves to an existing file under `docs/`
- [x] `search.json` diff is a pure deletion — 63 lines removed, 0 added — and byte formatting (2-space indent, no trailing newline) matches Quarto's own output
- [x] Reworded sentence confirmed in the rendered HTML
- [x] `docs/` committed alongside the source change


project-delphi (commented):
## Review

**Overview** — Two follow-ups from #127. Prunes 9 orphaned entries from `docs/search.json` that pointed at `posts/skills-vs-commands/skills-vs-commands.html` (the post's filename at commit `04899b5`, gone since), and rewords one date-anchored sentence in the post. Net −71/+8 across 4 files; the only hand-authored change is a single line of prose.

### Critical Issues

None.

### Important Issues

None. Both fixes verified independently:

- `search.json` parses, 508 entries, **every** `href` now resolves to a file that exists under `docs/`, no duplicate `objectID`s.
- Byte formatting matches Quarto's own output — 2-space indent, no trailing newline — so the diff is a pure deletion (63 lines removed, 0 added) with no reformatting noise.
- The prune survives a re-render: the targeted `quarto render` after pruning did not resurrect the entries, and it structurally can't — Quarto merges the index per rendered file, and the orphan has no source file to render from.
- No second copy of the dead URL elsewhere: `grep -rl` across `docs/` finds nothing, and this project generates no `sitemap.xml`.

### Suggestions

- **The pruning is the right call for this PR, but it leaves a latent repo problem worth naming separately.** `_freeze/` is gitignored and only 31 of 71 posts have a cache locally, so nobody can currently run `quarto render .` without executing ~40 posts against kernels (R, PySpark, HuggingFace) that may not exist in any one environment. That's fine until a change genuinely requires a whole-site rebuild — a nav change, a theme bump, `_quarto.yml`. Worth knowing which posts still execute cleanly before that day arrives, rather than discovering it under time pressure.
- The retracted paragraph still leads the indexed text for that section, because Quarto strips `<del>` when building the index. Unchanged from #127 and unavoidable in place — but note this PR halves the exposure rather than removing it: one indexed copy instead of two, and the surviving one is attached to a live URL where the reader immediately sees the strikethrough and the correction below it. That's the material improvement.
- Frontmatter `date` still unbumped across both PRs. Consistent with the previous decision, just noting the thread is still open.

### Strengths

- Chose the surgical fix over the destructive one and *documented the reasoning in the PR body*, including the freeze-cache count that justifies it. That's the difference between a shortcut and an argued trade-off.
- Verified the invariant that actually matters — every `href` resolving — rather than just eyeballing the diff.
- The reworded sentence is time-invariant: "for as long as the clerical work was yours to do" states the condition that changed instead of a date that will drift further from the byline every year.
- Generated-output diff stayed minimal; the only churn is one `data-listing-file-modified-sort` timestamp.

PR #129, opened 2026-07-29
build: make the site renderable off this laptop

project-delphi:
## Summary

Follow-up to the review on #128, which asked whether a full site render was safe. It wasn't — but not for the reason I gave there, and the investigation turned up something worse.

**The correction.** I claimed `quarto render .` would execute ~40 posts. Wrong in both directions:

- `freeze` is honoured only on a **project** render. A single-document render *always* executes. So the "safe, targeted" render documented in `CLAUDE.md` is the one that runs code; the full render is the one that uses the cache.
- Kernelspecs resolve while Quarto **indexes** the project, before `_freeze/` is consulted. One missing kernel fails every page, frozen or not — even prose posts.

**The real problem.** `_freeze/` was gitignored, and it was the only thing keeping 28 posts renderable. 11 of those are reproducible from source (venv + kernel + `requirements.txt`). The other 18 are not: no pinned kernel, no pinned versions, and dependencies (`diffusers`, `setfit`, `peft`, `transformers`, an `openai` client) whose environments no longer exist. A fresh clone or one `rm -rf` lost them silently.

Worse, 11 posts had **stale** frozen output: the cover-image commits edited their sources without re-rendering, and Quarto keys freeze on an md5 of the source file. A project render would have tried to execute exactly the posts that cannot execute.

## What changed

- **`_freeze/` is now tracked** (98 files, 8.5 MB, against a `docs/` already 28 MB). Scanned for credentials before staging — clean.
- **11 stale freeze hashes realigned.** For each, recovered the source revision matching the stored hash and asserted its code cells are byte-identical to HEAD's; only prose had changed. `pyspark` and `pinecone-vs-weaviate` left stale deliberately — no code cells, so nothing executes either way.
- **Full site rebuilt.** Zero cells executed. It also corrects real drift: five posts were published without the cover image their source has declared for months, and `search.json` drops 508 → 503, the five removed pointing at section anchors that no longer exist — the same stale-accumulation bug as the orphan pruned in #128, now cleared at source.
- **`make check-posts`** (`scripts/check_posts.py`, stdlib only) guards both invariants that broke: a code-executing post must pin a kernel and commit `requirements.txt`, and no post's frozen output may drift from its source. The 18 legacy posts are grandfathered so it's green on arrival and fires only on new drift.
- **`make kernels-stub`** registers the ten pinned kernel names against the base `.venv`, satisfying Quarto's index-time lookup with no ML dependencies. It skips names already registered — without that it repoints real kernels at the dependency-free venv and breaks every targeted render.
- **`CLAUDE.md` render guidance corrected**, including that a bare `quarto render .` can't see `--user` kernelspecs and deletes `docs/` before failing (`git checkout -- docs` to recover), plus a new section on the two tiers and why editing a legacy post breaks it.

## Test plan

- [x] `make check-posts` green; verified it fails correctly when a post loses its kernel pin, when `requirements.txt` is removed, and when a source is edited without re-rendering
- [x] `make kernels-stub` idempotent — keeps all 10 real kernels registered
- [x] Full render on this machine: exit 0, **0 cells executed**, 70 pages, 59 files changed, 0 deleted
- [x] **Fresh clone, isolated `JUPYTER_DATA_DIR`, zero ML dependencies**: `make check-posts` → green, `make kernels-stub` → 10 stubbed, `quarto render .` → exit 0, 0 cells executed, 70 pages
- [x] Clone output vs committed `docs/`: **zero missing files, one differing file** — homepage listing timestamps only. Byte-identical otherwise.
- [x] Rebuilt `search.json` has no orphan hrefs, confirming #128's manual prune agrees with a clean rebuild
- [x] Real kernels verified still pointing at their own venvs (`ripser`/`persim` importable) after testing


project-delphi (commented):
## Review

**Overview** — Tracks `_freeze/` (98 files, 8.5 MB), realigns 11 stale freeze hashes, rebuilds the site, and adds `make check-posts` + `make kernels-stub` with corrected `CLAUDE.md` guidance. 161 files, but only 5 are hand-authored: `.gitignore`, `CLAUDE.md`, `Makefile`, `scripts/check_posts.py`, and 98 freeze records + 59 rebuilt `docs/` pages that are generated output.

Three issues found during review, all fixed in `e355957` before this comment.

### Critical Issues

None outstanding.

### Fixed during review

- **`check_posts.py` passed when a legacy post's freeze record was deleted outright.** `check_freeze` returned early on `not record.exists()`, so the one genuinely unrecoverable state — a `LEGACY_NO_ENV` post losing the only reproducible copy of its output — was the one state it didn't catch. Verified by deleting `_freeze/posts/diffusers`: exit 0 before, exit 1 with a restore instruction after.
- **`STALE_FREEZE_OK` asserted a premise nothing enforced.** The exemption for `pyspark` and `pinecone-vs-weaviate` rests on them having no code cells; adding cells later would have silently suppressed real staleness. Now conditional on the claim still holding.
- **Two `D401` failures** against the repo's own `[tool.ruff]` (`select = ["D", "E", "F", "I"]`). `ruff check` and `black --check` now both clean.

### Verified

- `ipykernel` and `jupyter-client` are both in `uv.lock`, so `kernels-stub`'s `install` prerequisite (which prunes anything not in the lock) can't remove what the target then depends on.
- md5 use in `check_posts.py` matches Quarto's own freeze scheme and is not security-relevant; the repo's ruff config doesn't enable bandit (`S`) rules, so it won't trip lint.
- The `kernels-stub` clobbering bug (`b69804b`) was real and caught by running the target against a scratch clone, which repointed all ten `--user` kernels at the dependency-free venv. Real kernels were restored and re-verified (`ripser`/`persim` importable from `.venv-tda`).

### Suggestions

- `LEGACY_NO_ENV` is documented as shrink-only but nothing enforces the direction — a future edit could add a slug to silence the check rather than fix the post. A `git diff` assertion in the checker would be over-engineering; the comment plus review is probably the right weight for a single-author repo. Worth revisiting if it's ever silenced.
- The checker skips `.ipynb` posts entirely (`main()` continues when there's no `index.qmd`). Correct today, since Quarto doesn't execute notebooks by default, but an `execute: enabled: true` in a notebook's raw YAML cell would slip through unnoticed.
- The freeze-hash realignment in `0edc588` is a legitimate one-off — each was verified by recovering the source revision matching the stored hash and asserting code cells were byte-identical — but it's the kind of operation that should stay one-off. `check-posts` now makes the drift visible early enough that re-rendering is the normal fix.

### Strengths

- The end-to-end verification is the real content: a fresh clone with an isolated `JUPYTER_DATA_DIR` and zero ML dependencies renders all 70 pages, executes zero cells, and reproduces committed `docs/` with exactly one differing file — homepage listing timestamps. That proves the claim rather than asserting it.
- The rebuild surfaced two pre-existing bugs nobody was tracking: five posts published without the cover image their source declared, and five `search.json` entries pointing at anchors that no longer exist.
- Correcting `CLAUDE.md` in both directions matters more than the tooling. The old text said targeted renders were safe and full renders dangerous; the truth is the reverse on execution, and that error is what produced the workaround in #128.

PR #131, opened 2026-07-31
Explainability as a localization problem: MNIST attribution and Iris statistics

project-delphi:
A new long-form post, `posts/explainability-localization/`, plus its runnable source, figures and environment.

## The argument

Every explainability method answers one question — *which parts of the input, or of the model, are responsible for this output?* They differ on three axes:

1. **What counts as a part?** pixel, patch, conv channel, table column, variance direction.
2. **What counts as responsible?** local gradient, effect of removal, an axiomatic attribution, share of variance, effect of intervention.
3. **Does the answer come with error bars?** Almost never in ML. Almost always in statistics.

The third axis is the point of the piece. Part I (MNIST, black box) implements saliency, occlusion, Grad-CAM and integrated gradients in plain PyTorch; Part II (Iris, glass box) implements logistic coefficients with hand-derived Wald inference, one-way ANOVA, PCA and permutation importance. The synthesis places all eight on the grid and closes with what mechanistic interpretability inherits unchanged: point estimates.

## Three findings that fall out of the code

- **Integrated gradients assign exactly zero to all 617 baseline-valued pixels**, by construction — the attribution is a displacement times a gradient. Saliency puts its single largest gradient (1.35, the maximum of the whole map) on one of those same empty pixels. The two methods disagree for a definitional reason, not an empirical one.
- **Setosa is linearly separable, so the unpenalised MLE does not exist.** Over 100 Newton steps the coefficient norm climbs 1.67 → 92 and the largest standard error 1.37 → 1.4e7, neither converging. The most interpretable model in the book returns an explanation with infinite error bars — and the error bars are what tell you so.
- **Integrated gradients fails the Adebayo model-randomisation check**: magnitudes correlate at +0.99 with an untrained network's. The signed correlation is −0.01, so the information is entirely in the colours and none of it in the picture.

## Build notes worth reviewing

- The CNN ends in **global average pooling rather than a flatten**, which is what makes Grad-CAM's channel weighting valid. With a flatten head, all 49 cells of the pre-ReLU map came out negative and Grad-CAM returned an empty image. GAP costs accuracy (98.75%) and training time; the post states the trade-off rather than hiding it.
- **Training is on CPU, not MPS** — the accelerator is ~4x faster but not bit-reproducible across releases, and the prose names a specific test digit.
- **No drift between shown and executed code**: every code block is extracted from `src/` with `inspect.getsource` at render time. Verified after the fact by matching rendered blocks against the source files.
- **Nine cross-checks run during the render** and fail it on mismatch — IG completeness, Grad-CAM via hooks vs the head split, Grad-CAM == CAM/49, IG zero on baseline pixels, hand-derived logistic SEs vs `statsmodels`, ANOVA vs `scipy`, PCA vs `scikit-learn`, separability by linear program, permutation importance vs `sklearn.inspection`. Their output is reproduced in a collapsible callout at the top of the post.
- Ships **light-only** (`color-scheme: light`), because the post asks the reader to check specific colours by eye and Chrome's auto dark mode inverts the figure PNGs.

## Repo housekeeping

- New `.venv-explainability` + `explainability-blog` kernel, with a curated pinned `requirements.txt`.
- `explainability-blog` added to the root `Makefile`'s `kernels-stub` list so a fresh clone can render the project from `_freeze/` without the ML dependencies.
- `docs/` re-rendered in the same commit. `make check-posts` is green.

## Verification

- `make figures` — 12 figures regenerate standalone, all gates pass.
- `quarto render posts/explainability-localization/index.qmd` — builds clean.
- `make check-posts` — green before and after the project render.
- `quarto render .` — full site render; only `docs/index.html`, `listings.json` and `search.json` change outside the new post, so no other post's output drifted.
- Rendered page read end to end; every figure referenced in prose exists and shows what the text claims.


project-delphi:
## Self-review

Reviewed the diff adversarially rather than re-confirming the authoring. Three real defects, now fixed in 5044915.

### 1. The opening paragraph stated the wrong rank — Critical

> The network's probability for the correct answer is 0.00055 — not second place, but **fifth, behind an 8 and a 2**.

It ranks **third**, behind only the predicted 6 and an 8 — and it is *ahead* of the 2. The ranking is `[6, 8, 5, 2, 0, 9, 1, 4, 3, 7]`.

The number was right for an earlier draft, before the architecture changed from a flatten head to global average pooling and the running example moved from "an 8 read as a 0" to "a 5 read as a 6". The Figure 1 caption inherited the same error.

This is the failure the post's whole "every number is computed at build time" framing exists to prevent, and it landed anyway because that one sentence was *prose* rather than an interpolation — nothing checked it. Rank and the list of classes ahead of it are now derived at render time, so the sentence cannot go stale again.

### 2. Runner-up class was tie-broken off the clipped array — Minor

`fig_examples` ranked with `np.argsort(np.clip(probs, 1e-6, 1.0))`. For the control digit **nine of ten classes sit below the 1e-6 display floor**, so argsort returned an arbitrary member of that tie: the panel reported `p(9) = 8.0e-08` when the real runner-up was `p(3) = 5.9e-07`. Now ranks on the raw probabilities. Wrong only on the control panel, which is why it survived the first read.

### 3. An ungated prose claim — Minor, but the same class of bug as (1)

> the single largest gradient in the entire map, 1.35, sits on one of them

True for this checkpoint, but it was the only numeric claim in the post not backed by an assertion — a re-render that moved the maximum onto the stroke would have falsified it silently. Now a build gate: the render fails and the sentence has to be rewritten.

### Not changed, and why

- **`randomised_copy(model, seed)` ignores its `model` argument.** It constructs `SmallCNN()` directly, so the parameter is decorative. Harmless here since there is one architecture, and worth leaving as the signature a caller would expect.
- **`separation_path`'s docstring mentions a column that no longer exists** ("the fitted probability closest to a mistake"), left over from trimming the frame. Cosmetic; noting it rather than churning the diff again.
- **`ruff` and `black` do not pass on `src/`.** Neither do they on `posts/bayesian-bootstrap/src/` (79 errors), and CLAUDE.md says the hooks are not installed by default. The `E`/`F`/`I` classes *are* clean after fixing a genuinely unused import and four unsorted import blocks; the remainder is `D`-rule noise from `pyproject.toml` selecting numpy-convention docstrings while `[tool.pydoclint]` asks for Google style. That contradiction predates this PR.

### Verified after the fixes

`make figures` (14 assertions pass) → single-doc render → `make check-posts` → full project render → `make check-posts`. Outside the new post, only `docs/index.html`, `listings.json` and `search.json` changed, so no other post's output drifted.

PR #132, opened 2026-08-01
build: shared Hugging Face environments and kernel plumbing

project-delphi:
## Why

Thirteen posts on this blog date from March–April 2025 and were written against `transformers` 4.4x, `datasets` 2.x, `optimum` 1.x, `langchain` 0.1 and `openai` 0.x. Their code no longer runs. All thirteen are grandfathered in `LEGACY_NO_ENV` — no venv, no pinned kernel, no `requirements.txt` — so their committed `_freeze/` record is the only surviving copy of what they compute.

Quarto keys frozen output on an md5 of the source, so *any* edit to one of these posts invalidates its record and makes the next project render try to execute a post with no environment. Fixing the code and building the environment are therefore the same task. This PR is the environment half, split out so it can be reviewed on its own.

**No `posts/` source changes**, so every `_freeze/` record stays valid and `make check-posts` stays green.

## What

Two shared venvs rather than thirteen per-post ones. The thirteen share ~90% of their dependency closure, almost all of it torch; thirteen `.venv-<slug>` directories would be 20 GB+ of near-identical wheels.

| venv | kernel | `transformers` | posts |
|---|---|---|---|
| `.venv-huggingface` | `huggingface-blog` | 5.14.1 | 11 |
| `.venv-huggingface-t4` | `huggingface-t4-blog` | 4.57.6 | `optimum`, `setfit` |

## Why the split

Two laggard packages cap `transformers`, and neither is worth holding the other eleven posts back for:

- **`optimum-onnx` 0.1.0** declares `transformers<4.58.0`. This package is also *why* the `optimum` post is broken at all: `optimum` 2.x moved ONNX support out of the core package, so the post's `pip install optimum` line no longer provides `optimum.onnxruntime`. (The post's import statements themselves turn out to be fine — verified that `optimum-onnx` still registers under `optimum.onnxruntime` and `optimum.exporters.onnx`, with `main_export`'s signature unchanged.)
- **`setfit` 1.1.3** declares no upper bound at all, but imports `transformers.training_args.default_logdir`, which 5.x removed.

That second one is worth calling out: `uv pip install --dry-run` reported a **clean resolve** for a combination that cannot import. CLAUDE.md now says to import the packages rather than trust a resolve, because a resolver-only check does not catch this.

## Changes

- `Makefile` — register both kernels in `kernels-stub`, so a fresh clone can render frozen output without the ML stack.
- `CLAUDE.md` — document the shared-venv exception under § Architecture, including the trade it makes: `requirements.txt` for these thirteen is a copy of a shared freeze, so bumping one post's dependency changes the recorded versions for every post on that kernel.

## Verification

- `make check-posts` → `OK (18 legacy posts grandfathered)`
- `make kernels-stub` → `kept huggingface-blog (already registered)`, likewise `huggingface-t4-blog` — confirms it does not repoint real kernels at the dependency-free `.venv`
- Both venvs import-tested, not just resolved. MPS available.

## What comes next

Four more PRs, each shrinking `LEGACY_NO_ENV`: (2) `datasets`, `hub`, `tokenizers`, `peft` · (3) `transformers-library`, `hugging-face-evaluate-library`, `optimum`, `setfit` · (4) `diffusers` · (5) the four adjacent posts.


project-delphi:
## Review

Scope: 2 files, +9/-1 as opened. Documentation and one `Makefile` line — no post sources, so `_freeze/` integrity is not at risk here.

### Overview

Adds two shared venvs + kernels for the thirteen HuggingFace-era posts and registers them in `kernels-stub`. Correctly sequenced: environment first, post edits later, so `make check-posts` stays green throughout rather than going red the moment a source is touched.

### Correctness

- ✅ `kernels-stub` change is right. Verified it does not repoint the real kernels at the dependency-free `.venv` — `make kernels-stub` reports `kept huggingface-blog (already registered)` for both. That guard was the whole point of the existing loop and the addition respects it.
- ✅ Post counts are consistent: 11 + 2 = 13, and the enumerated list has 13 entries.
- ✅ The `setfit` finding is verified, not inferred. `uv pip install --dry-run` reports a clean resolve against transformers 5; the import then fails on `transformers.training_args.default_logdir`. Worth having in CLAUDE.md.

### Issues found and fixed in 3bcf4e8

1. **No way to rebuild the venvs** (the significant one). CLAUDE.md named `.venv-huggingface` and `.venv-huggingface-t4` but nothing in the repo recorded their package sets — those are gitignored, and `requirements.txt` does not arrive until PR 2. A clone could stub the kernels but not build the real ones. In a PR whose entire purpose is reproducibility, that is the wrong gap to leave open for four PRs. Package sets are now recorded inline.

2. **`20 GB+` was asserted, not measured.** Actual: 1.2 GB per venv, and `uv` copies rather than hardlinks here (verified link count 1), so thirteen would be ~15 GB. Conclusion unchanged, figure corrected. CLAUDE.md is load-bearing for future work; a number in it should be one someone checked.

3. **Introduced a contradiction.** The pre-existing "Two tiers" paragraph says these posts' dependencies have environments that "no longer exist anywhere" — false for thirteen of them as of this PR. Reworded, with `LEGACY_NO_ENV`'s size flagged as the live progress measure.

4. **Environment did not match its own documentation.** `setfit` was still installed in `.venv-huggingface` from the first install attempt — present but broken at import, and would have been captured in eleven posts' `requirements.txt` in PR 2. Removed.

### Risks accepted

- The two venvs drift independently of `uv.lock`, which covers only the base toolchain. Mitigated by per-post `requirements.txt` in later PRs, and CLAUDE.md now states the consequence (bumping one post's dependency changes recorded versions for every post on that kernel).
- `transformers` 4.57.6 for `optimum`/`setfit` is a deliberate lag on two upstream packages, not an oversight. Revisit when `optimum-onnx` moves past `transformers<4.58.0`.

### Not applicable

No tests (this repo's test is `make check-posts`, green). No security surface — no network calls, credentials, or user input in the diff.

**Verdict: nothing Critical.** The four issues above are fixed in `3bcf4e8`.

PR #133, opened 2026-08-01
fix(posts): repair and rewrite datasets, hub, tokenizers and peft

project-delphi:
Stacked on #132 — merge that first, and this diff will rebase onto `main` cleanly.

## The bug, and why it was invisible

These four posts ran against `datasets` 2.x and `huggingface_hub` 0.2x. Under current libraries the dataset loads are hard failures:

```
HfUriError: Invalid HF URI 'hf://datasets/imdb@...'.
Repository id must be 'namespace/name', got 'imdb'.
```

`datasets` 5.x requires every id to be `namespace/name`; the bare canonical aliases are gone.

**This does not reproduce on the machine that wrote the posts.** `datasets` silently falls back to a stale local cache, logging `Using the latest cached version of the dataset since imdb couldn't be found on the Hugging Face Hub`. The cells pass locally and fail for every reader. I only caught it by re-running against a clean `HF_HOME`, which is now how I check the remaining posts. `imdb`, `ag_news` and `glue` are all affected — the last two reach `setfit` and `transformers-library` in PR 3.

## Per post

Each now pins `jupyter: huggingface-blog`, carries `requirements.txt`, and leaves `LEGACY_NO_ENV` (18 → 14). All rewritten to the house conventions around a single argument.

**`datasets`** — *`load_dataset` memory-maps rather than reads.* Row 24,999 costs the same 0.05 ms as row 0; splitting 25,000 rows takes 2 ms.

> An earlier draft tried to demonstrate this with RSS and it did not survive review. Memory-mapped pages count toward resident memory once touched, so the delta came out +141 MB one run and +42 MB the next — and peak RSS (373 MB) actually exceeded the on-disk size, reading as a *refutation* of the claim. Access time is the measurement that distinguishes a memory-map from a read; RSS is not. The post now says so explicitly.

**`hub`** — *every repo is a git repo, so an unpinned `from_pretrained` is a dependency on a moving target.* Shows commit history via `list_repo_commits`, pins a `revision`, and uses the August 2024 `runwayml/stable-diffusion-v1-5` deletion as the case where pinning still does not save you.

**`tokenizers`** — *the vocabulary is learned, and too little text fails visibly.* The post's own claimed output was false: it promised `['[CLS]','hello','world','[SEP]']`, but its 341-byte corpus actually yields `['[UNK]','e','l','l','o','wor','l','d']`. That bug is now the subject — the trainer returns a 187-token vocabulary when asked for 1,000 (it ran out of corpus), and `[UNK]` on a capital `H` means that byte was never observed. Trained on 25k IMDB reviews instead, identical code gives `['Hel','lo','world']`.

**`peft`** — *LoRA works because the update is low-rank, not because 0.24% of parameters happens to suffice.* Adds the measurement that makes it concrete: a **1.13 MB adapter against a 522.7 MB base, 461× smaller**, and `merge_and_unload()` returning a plain `GPT2LMHeadModel` at exactly the original parameter count.

## Also fixed

- Trailing generation scaffolding — `datasets` ended with "Let me know if you need more examples or modifications! 🚀", `peft` with "Let me know if you want to add plots".
- `peft`'s comparison table was tab-separated text that markdown never rendered as a table.
- `tokenizers` had a `python` fence with prose inside it, and a cell that built a tokenizer and produced no output.
- Dead cells that constructed configs and printed nothing now print something.

## Verification

- `make check-posts` → `OK (14 legacy posts grandfathered)`
- All four rendered with `QUARTO_PYTHON=.venv-huggingface/bin/python`; `cell-output-error` count is 0 in each
- Every number quoted in prose is read off the committed render, not asserted

> `check-posts` earned its keep here: it caught a one-word wording fix I made to `datasets` *after* rendering, exactly the stale-freeze failure mode it was written for.


project-delphi:
## Review

Scope: 4 posts rewritten, 4 `requirements.txt` added, `LEGACY_NO_ENV` 18 → 14, plus re-rendered `docs/` and `_freeze/`.

### Verified, not assumed

- **The `HfUriError` claim reproduces.** Ran `imdb`, `ag_news` and `glue` against a clean `HF_HOME`: all three raise. Also confirmed the masking behaviour — with the local cache present they silently succeed while logging `couldn't be found on the Hugging Face Hub`. This is the correct diagnosis of "the code doesn't run".
- **Every number in prose matches the committed render.** Spot-checked all four: `0.05 ms` / `1.9 ms` (datasets), the commit shas and 12-file snapshot (hub), `187` vocab and `['[UNK]','e','l','l','o',...]` (tokenizers), `1.13 MB` / `522.7 MB` / `461x` (peft).
- **Cross-links resolve on disk**, not just in source: `docs/posts/tokenizers/index.html` → `../../posts/datasets/index.html` exists.
- **`requirements.txt` matches the pinned kernel** in all four (`transformers==5.14.1` under `huggingface-blog`), and none of them carries the broken `setfit`.
- `cell-output-error` count is 0 in all four rendered pages.

### Judgement calls worth flagging to a reviewer

- **The RSS demonstration was cut, not fixed.** The first draft measured memory to prove memory-mapping and the measurement contradicted the claim (peak RSS 373 MB against 127 MB on disk; deltas of +141 MB and +42 MB on consecutive runs). Rather than keep a noisy number that argued the opposite, the post now measures access time and says explicitly why RSS is the wrong instrument. Worth a look if you disagree with dropping it.
- **`tokenizers` now trains on 25k IMDB reviews** where it previously used a 341-byte file. That adds a few seconds to the render and a dataset dependency to a post that had none — deliberate, since the small-corpus failure is the post's subject and it needs the contrast.
- **`peft` downloads GPT-2 twice** (once for LoRA, once for prompt tuning) because `get_peft_model` mutates in place. Cached after the first render, but it is a real cost.

### Nothing Critical

No correctness defects found. `make check-posts` → `OK (14 legacy posts grandfathered)`.

> The checker caught one genuine mistake during this work: a one-word wording fix applied to `datasets` *after* rendering left the freeze hash stale, and `check-posts` failed exactly as designed.

PR #134, opened 2026-08-01
fix(posts): repair transformers-library, evaluate, optimum and setfit

project-delphi:
Stacked on #133 (which is stacked on #132). `LEGACY_NO_ENV` 14 → 10.

## `transformers` 5 removed more than I expected

All confirmed by running the code, not by reading changelogs:

| Was | Now | Symptom |
|---|---|---|
| `TrainingArguments(evaluation_strategy=)` | `eval_strategy` | `TypeError` |
| `Trainer(tokenizer=)` | `processing_class` | removed |
| `pipeline(..., grouped_entities=True)` | `aggregation_strategy` | removed |
| `pipeline("summarization")` | — | **`KeyError`** |
| `pipeline("translation_en_to_fr")` | — | **`KeyError`** |
| `pipeline("text2text-generation")` | — | **`KeyError`** |

The last three are the surprise: the sequence-to-sequence pipeline tasks are **no longer registered at all**. Both `transformers-library` and the `evaluate` post depended on them. Seq2seq models are now driven directly via `AutoModelForSeq2SeqLM` + `generate()`, writing the task prefix (`"summarize: "`) yourself — which the deleted pipeline used to hold for you.

## `optimum`'s code was fine; its install line was not

`optimum` 2.x moved every ONNX backend into a separate `optimum-onnx` distribution, so `pip install optimum[onnxruntime]` yields a package with no `optimum.onnxruntime` in it. The **import paths are unchanged**, so the failure lands on a line that is correct and looks like a code bug.

That package pins `transformers<4.58`, which is why `optimum` and `setfit` sit on the `huggingface-t4-blog` kernel from #132 while the rest run transformers 5.

## Three silent wrongnesses — none of which raised

1. **`transformers-library` trained on one class.** It used `split="train[:1%]"`, and IMDB stores all 12,500 negative reviews before all 12,500 positive ones. Evaluation reported **100% accuracy** for a model that always answered "negative". Shuffled, it reports **0.82** — and that number is real.
2. **`optimum` benchmarked a random model.** It exported `bert-base-uncased` with `task="text-classification"`, which attaches a *randomly initialised* head, then printed its confident prediction as if it meant something.
3. **`optimum`'s INT8 section never quantized.** It loaded an ONNX model and called `onnx.save` on it unmodified, then announced "Quantized model loaded successfully". Replaced with a real `ORTQuantizer` pass.

The `evaluate` post also shipped a commented-out BERTScore cell annotated *"There's a small bug here, I don't have time to fix it - however similar code should work."* There was no bug — `bert_score` was not installed. It is now in the venv and the cell runs.

## What each post now argues

- **`transformers-library`** — *`pipeline()` is a tokenizer, a model and a post-processor.* Reconstructs a pipeline's output by hand from its own attributes and matches it to seven decimals. Documents that transformers 5 auto-places models on MPS, so hand-written forward passes need `.to(model.device)` — I hit this as an uninformative `SIGBUS` with no traceback before finding the real error underneath.
- **`evaluate`** — *the same sentence pair scores **0.0** (ROUGE-2 and BLEU) and **0.96** (BERTScore).* One uniform interface over incompatible theories of correctness. The custom-metric section now demonstrates something: two prediction sets that accuracy scores identically at 0.75, separated 5× by a cost-weighted metric.
- **`optimum`** — *ONNX export is a compile step; measure the trade.* **3.24×** CPU latency, **4× smaller** with INT8, predictions intact to six decimals — with an explicit caveat that this is batch-1 CPU on one machine and not a quotable constant.
- **`setfit`** — *identical code and 16 examples scores **+0.210** over baseline on SST-2 and **−0.095** on MRPC.* Few-shot works when the frozen encoder already separates the classes; on MRPC it loses to always guessing the majority class.

The `setfit` post deliberately keeps a **wrong** prediction in its output — the model labels "a tedious, joyless slog" as positive. That is what 0.71 accuracy looks like, and cherry-picking examples it gets right would misrepresent the method.

## Also

`Trainer` and SetFit write checkpoints relative to the render cwd, so rendering these posts dropped **1.3 GB** of weights into `posts/` and the repo root. Removed and gitignored.

## Verification

- `make check-posts` → `OK (10 legacy posts grandfathered)`
- All four rendered against their pinned kernels; `cell-output-error` count 0 in each
- Every figure quoted above is read off the committed render


project-delphi:
## Review

Scope: 4 posts, `LEGACY_NO_ENV` 14 → 10, two of them moved onto the `huggingface-t4-blog` kernel.

### Verified

- **Every removal claim was reproduced by running it**, not read from a changelog. `evaluation_strategy` raises `TypeError`; `grouped_entities` is absent from `TokenClassificationPipeline._sanitize_parameters`; `Trainer.__init__` has `processing_class` and no `tokenizer`; and `pipeline("summarization" | "translation" | "text2text-generation")` all raise `KeyError` against the registered-task list.
- **The one-class training bug is real and quantified.** `Counter(load_dataset("stanfordnlp/imdb", split="train[:2%]")["label"])` → `{0: 500}`. Shuffled → `{1: 254, 0: 246}`. The post now renders both, so the trap is visible rather than asserted.
- **`requirements.txt` matches each kernel**: `transformers==4.57.6` for `optimum`/`setfit`, `5.14.1` for the rest. Both t4 posts carry `optimum-onnx` and `setfit`, which is expected from a shared venv.
- All four render with `cell-output-error` = 0; every quoted figure (3.24×, 255.5→64.2 MB, +0.210/−0.095, 0.0 vs 0.96) is read off the committed HTML.

### Things a reviewer should push on

- **`setfit` publishes a model that is wrong in its own demo** — "a tedious, joyless slog" classified positive. This is deliberate and argued for in the text, but it is the kind of choice worth disagreeing with. The alternative was picking examples it gets right, which would misrepresent a 0.71-accuracy model.
- **`optimum`'s benchmark is batch-1, CPU, one machine.** The post says so and declines to present the ratio as a constant, but a 3.24× number in a heading will get quoted out of context regardless.
- **The `evaluate` custom-metric section was rewritten mid-review.** The first version returned 0.75 — *identical* to plain accuracy — so it demonstrated nothing. It now contrasts two prediction sets that accuracy scores the same and the cost metric separates 5×.
- **`posts/optimum/onnx/bert/` was deleted.** Those were committed config/tokenizer files for an export the post no longer performs (it now exports to a temp dir, so no 255 MB artefact lands in the repo). Intentional, but it is a deletion of tracked content.

### Follow-up not in this PR

Rendering these posts wrote 1.3 GB of `Trainer`/SetFit checkpoints into `posts/` and the repo root. Removed and gitignored here — but worth knowing that any future fine-tuning post will do the same unless `output_dir` is pointed outside the tree.

### Nothing Critical

`make check-posts` → `OK (10 legacy posts grandfathered)`.

PR #135, opened 2026-08-01
fix(posts): repair and rewrite diffusers, executed against SD 1.5

project-delphi:
Stacked on #134. `LEGACY_NO_ENV` 10 → 9. **All five images regenerated on MPS**, not reused from the old commit.

## Dead repo ids

`runwayml/stable-diffusion-v1-5` and `runwayml/stable-diffusion-inpainting` were deleted by their owner in August 2024 and survive only as redirects. Now addressed under `stable-diffusion-v1-5/`.

## Three bugs that never raised

- **`ipe = pipe.to("mps")`** — a typo binding a stray name, so every generation in this post ran on CPU.
- **The inpainting mask was opened as `RGB`.** It is a bilevel image and wants `"L"`.
- **`torch_dtype=float16` on MPS** — a known black-image failure mode.

## Two more that only running it could reveal

This is the argument for executing the post rather than shipping stored images.

**1. ControlNet was producing garbage.** It conditioned on a stored 225×225 canny map upscaled to 512. That does not degrade gracefully: upscaling one-pixel lines produces broken *dotted* lines, and the model faithfully reproduced the dots as content — returning a grey wash with edge lines drawn on top rather than a picture. The first render of this PR shipped exactly that, and I only caught it by looking at the image.

Fixed by computing Canny from `base.png` at generation resolution with `cv2`. The output is now a winter scene holding the source composition exactly — dome, framing pines, road curve — which is the actual ControlNet argument. `canny_edges.png` is deleted, having become unused.

**2. The inpainting prompt described a different image.** It asked for *"a black cat with glowing eyes"* against a photograph of Half Dome, and produced a **castle**. The prompt was inherited from some earlier base image and had nothing to do with this one. The model fills a masked region plausibly whether or not your prompt suits its surroundings, and gives no signal when the two disagree. Reprompted to match; the mismatch is now written up as the caveat it always was.

## What it argues now

A pipeline is four separable parts, and every way of controlling it intervenes on exactly one:

| Technique | Component |
|---|---|
| Prompt | text encoder |
| Sampler swap | scheduler — **0 parameters**, hence freely swappable |
| Inpainting mask | which latents may move |
| ControlNet | a second network beside the UNet |

The measured split is printed in the post: text encoder 123.1M, UNet 859.5M, VAE 83.7M, scheduler 0.0M.

The scheduler section prints a **mean absolute pixel difference** (14.9/255) rather than asserting the two images differ — at 25 steps on a converged prompt they look nearly identical, and "compare them yourself" would have been evasive.

## Data provenance

The post had none. It now covers LAION-5B: what was scraped, by whom, why (an explicit response to DALL·E's undisclosed corpus), and the December 2023 Stanford Internet Observatory CSAM finding that led to the dataset's withdrawal and the filtered Re-LAION release — all upstream of these weights.

## Requirements churn

`torchvision` and `opencv-python-headless` join the shared venv, so **all seven** `huggingface-blog` posts get refreshed `requirements.txt`. That is the documented consequence of a shared environment (CLAUDE.md § Architecture), not incidental noise.

## Verification

- `make check-posts` → `OK (9 legacy posts grandfathered)`
- 6/6 cells executed, `cell-output-error` = 0
- Every generated image inspected visually, not just checked for non-blackness — which is how both failures above were found


project-delphi:
## Review

Scope: one post, executed for real against ~9.5 GB of Stable Diffusion weights. All five images regenerated.

### Verified

- **Component parameter counts are read off the render**, not asserted: text encoder 123.1M, UNet 859.5M, VAE 83.7M, scheduler **0.0M** — the last being the post's justification for calling the scheduler freely swappable.
- **The scheduler claim is now measured** (mean |Δpixel| 14.9/255) rather than asking the reader to eyeball two similar images. Earlier drafts said "not identical" and left it there, which was evasive given how close they look.
- Both regenerated figures inspected visually. The ControlNet triptych shows source → edge map → winter scene holding the same composition; the inpainting shows a castle on the summit with the framing pines pixel-identical.
- 6/6 cells, `cell-output-error` = 0.

### The two failures this PR fixes are the argument for executing it

Both were invisible to every automated check — no exception, no warning, non-black output, correct image dimensions:

1. **ControlNet was returning a tracing of its own control image**, because a 225×225 canny map upscaled to 512 becomes dotted lines that the model reproduces as content. The first render of this branch shipped exactly that, and only looking at the PNG caught it.
2. **The inpainting prompt described a different photograph.** It asked for a black cat and got a castle, and would have kept doing so indefinitely.

A reviewer should take that as the general lesson: for generative posts, "the cell ran" is close to no evidence at all.

### Worth pushing back on

- **`safety_checker=None`** is set on all three pipelines. Justified in the text (the checker would make committed figures depend on its verdict) but it is a deliberate disabling of a safety component in published code, and readers copy published code. The post says to leave it enabled for anything user-facing; you may want that stronger.
- **`opencv-python-headless` and `torchvision` join the shared venv** for one post's benefit, refreshing `requirements.txt` for all seven `huggingface-blog` posts. Documented as the shared-venv trade, but it is real coupling.
- **Re-rendering this post costs ~9.5 GB of weights and several minutes of MPS time.** That is now the standing cost of touching one word of its prose.

### Nothing Critical

`make check-posts` → `OK (9 legacy posts grandfathered)`.

PR #136, opened 2026-08-01
fix(posts): repair the four adjacent posts; LEGACY_NO_ENV down to 5

project-delphi:
Final PR of the series, stacked on #135. **`LEGACY_NO_ENV` 9 → 5.** All thirteen HuggingFace-era posts now pin a kernel, carry `requirements.txt`, and execute against current libraries.

## `langchain` — deleted, not deprecated

LangChain 1.3 does not deprecate the 0.1 API, it removes the modules:

```
GONE  langchain.prompts.PromptTemplate                     (ModuleNotFoundError)
GONE  langchain.memory.ConversationBufferMemory            (ModuleNotFoundError)
GONE  langchain.chains.LLMChain                            (ModuleNotFoundError)
GONE  langchain.text_splitter.RecursiveCharacterTextSplitter (ModuleNotFoundError)
```

`ConversationBufferMemory`, `LLMChain`, `ChatOpenAI` and `.run()` have nowhere left to import from. Only one of the post's six code blocks was ever executed, and it was one of the casualties.

Rewritten to LCEL over a local **Qwen2.5-0.5B-Instruct** — no API key, no cost, reproducible from `requirements.txt`. Memory is now *demonstrated* to be a wrapper rather than a model property: the same chain answers **"Your name is Ravi."** with history replayed and **"I do not have a personal name."** without it.

Two wrappers are needed and both matter — `ChatHuggingFace` to apply the chat template, and `return_full_text=False`, without which every reply arrives with the entire templated conversation glued to the front.

> Worth flagging: `RunnableWithMessageHistory` — the replacement — emits **its own** deprecation warning pointing at LangGraph. The post says so rather than hiding it, because that is its closing argument about churn arriving on schedule.

## `classifying_text_chunks` — never produced a result

All five cells only *defined* functions. Nothing was ever called, so the post showed the reader no output at all — and `PyPDF2`, `langchain.text_splitter` and `openai.ChatCompletion.create` are all gone besides.

It now generates synthetic review PDFs with matplotlib, then extracts, chunks, embeds and classifies end to end. (`pdf.fonttype = 42` is load-bearing: matplotlib's default renders text as vector paths and extraction silently returns nothing.)

Making it run surfaced the bug it was built on and never noticed. It aggregated chunk labels with `int(round(np.mean(preds)))` over an **ordinal** label set:

```
chunk predictions : ['good', 'bad', 'bad']
mean, then round  : neutral      <-- negative chunks outnumber positive 2:1
majority vote     : bad
disagreement (sd) : 0.94
```

`good`=2 and `bad`=0 average to the middle, so a polarised document is indistinguishable from a lukewarm one. That aggregation step is now the post's subject, and the FastAPI endpoint returns the distribution and a `mixed` flag instead of destroying the disagreement server-side.

## The two that already ran

Both were catalogues with nothing at stake. Both now demonstrate something.

**`evaluation-metrics`** — real data (Wisconsin Diagnostic Breast Cancer) and a genuine asymmetry between the two errors:

```
model                 acc   prec  recall     f1  missed cancers
always benign       0.626  0.000   0.000  0.000              64
logistic            0.953  0.938   0.938  0.938               4
cautious (p>0.9)    0.953  1.000   0.875  0.933               8
AUC (both models): 0.9917
```

Identical accuracy, identical AUC, **twice as many missed cancers** — and precision *prefers* the worse one. Plus a threshold sweep making the trade explicit, and MAE/RMSE ranking two regressors in opposite orders.

**`soft-vs-hard-prompts`** — the old demo regressed a random tensor toward another random tensor and demonstrated nothing about prompts. Now a real `peft` prompt-tuning run initialised *from* the hard prompt, so step zero is exactly the text:

| | nearest vocabulary tokens | mean cosine |
|---|---|---|
| at init | `['Class','ify',' the',' review',' sentiment',':',...]` | **1.000** |
| after 60 steps | `[' mascot','ify',' the',' sitcom','龍�',':',...]` | **0.333** |

The prompt provably leaves the vocabulary — the soft/hard distinction made measurable rather than asserted.

## Verification

- `make check-posts` → `OK (5 legacy posts grandfathered)` — down from 18 at the start of the series
- **Full project render clean across all 73 pages**; frozen output held for every untouched post
- `grep -rl cell-output-error docs/posts/*/index.html` → no matches anywhere on the site
- All 13 posts confirmed pinning a kernel and carrying `requirements.txt`

`LEGACY_NO_ENV` now contains only the five genuinely orphaned posts: `data-types`, `features-importance-after-clustering`, `poor-persons-bayesian`, `post-with-code`, `working-with-quarto`.


project-delphi:
## Review

Final PR of the series. Scope: 4 posts, `LEGACY_NO_ENV` 9 → 5, plus a full-site render.

### Verified

- **Every deletion claim is produced by the post itself at render time** — the `langchain` post prints its own four `ModuleNotFoundError`s rather than describing them, so the evidence cannot go stale silently.
- **The memory contrast is a true controlled comparison**: same chain, same model object, same process; only `history` differs. "Your name is Ravi." vs "I do not have a personal name."
- **The ordinal-aggregation bug reproduces in the rendered output** — `['good','bad','bad']` → `neutral`, with majority vote saying `bad` on identical predictions.
- **The metrics table shows the disagreement it claims**: 0.953/0.953 accuracy, 0.9917/0.9917 AUC, 4 vs 8 missed cancers.
- **The prompt-drift measurement is exact** because the soft prompt is initialised *from* the hard prompt, so cosine 1.000 at step 0 is a fact about the setup rather than a coincidence.
- Full project render clean across **73 pages**; `grep -rl cell-output-error docs/posts/*/index.html` returns nothing site-wide.

### Judgement calls to challenge

- **`RunnableWithMessageHistory` is itself deprecated**, and the post teaches it anyway with the warning quoted inline. The alternative was pulling in LangGraph, which is a much larger dependency for a post about composition primitives. Reasonable people could prefer the other choice.
- **`evaluation-metrics` and `soft-vs-hard-prompts` were not broken.** Their code ran fine before. Everything here is prose plus a new demonstration, which is scope you approved but is worth seeing clearly as a *rewrite* rather than a repair.
- **Six synthetic documents** underpin `classifying_text_chunks`. The post says explicitly that this is not an evaluation and that the aggregation argument does not depend on the classifier being good — but the toy corpus is doing real work in making the mixed-document case exact.
- **A wrong prediction is shipped on purpose** in `soft-vs-hard-prompts`' sibling post (`setfit`, #134) and the tone here is consistent: these posts now show failures rather than curating around them.

### Series outcome

`LEGACY_NO_ENV` went 18 → 5 across five PRs, and the five that remain are genuinely orphaned posts outside this scope. The invariant `check_posts.py` protects held at every step — including catching one stale freeze hash of mine in #133.

### Nothing Critical

`make check-posts` → `OK (5 legacy posts grandfathered)`.

PR #137, opened 2026-08-02
feat(posts): LLM agents from first principles, executed against Groq

project-delphi:
## Summary

- New post `posts/llm-agents-from-first-principles/`. The blog's three existing agent posts (`langgraph`, `langgraph-vs-llamaindex`, `how-does-rag-work`) all explain agents *through* a framework. This one removes the framework to answer the question underneath: mechanically, an agent is deterministic control flow wrapped around exactly one stochastic call.
- `agent.py` is the whole agent in one **standard-library-only** file — `Transcript` (dialogue state), `AgentState` (scaffold state), `Tool`/`ToolRegistry`, `parse()`, `render()`, `Agent.run()`. No `requests`, no SDK; the zero-dependency property is the argument.
- **Both worked examples execute live at render time** against Groq's free tier (`llama-3.3-70b-versatile`). Nothing in the post is a hand-typed transcript. Example 1 terminates in one turn with no tool call; Example 2 does two genuine round trips (`calculator` → `is_prime` → `FINAL`).
- Two things the post documents because they actually happened while writing it: Cloudflare 1010-rejects the stdlib's default `Python-urllib` user agent, and Llama 3.3 emits `ACTION: None` next to a valid `FINAL:` line, which the first version of the parser read as a malformed tool call and burned four of six turns on.
- `Makefile`: adds `llm-agents` to `kernels-stub`. Without it a fresh clone cannot render *any* page — Quarto resolves kernelspecs while indexing the project, before it consults `_freeze/`.
- `CLAUDE.md`: records that this post needs `GROQ_API_KEY` only for a *single-document* render (a project render reuses the committed freeze), and that its captured output is a sample from a stochastic policy rather than a reproducible fixed point — so it should not be re-rendered to "refresh" anything.

Note on the brief this was written from: it specified `(128^2 + 7) / 13` as the tool-use example, but that is 1260.846…, so the primality question is vacuous. Changed to `(128^2 + 3) / 7 = 2341`, which is exactly integral and genuinely prime.

## Test plan

- [x] `agent.py` runs standalone against live Groq (`python agent.py`) — both examples produce the intended trajectories
- [x] `quarto render posts/llm-agents-from-first-principles/index.qmd` executes cleanly end-to-end with a live key
- [x] Rendered transcripts verified against the prose; arithmetic checked independently (`(37·4−9)/5 = 27.8`, `128²+3 = 16387`, `16387/7 = 2341`, 2341 prime)
- [x] Trajectories confirmed stable across two separate renders
- [x] `make check-posts` → OK (5 legacy posts grandfathered); post pins a kernel and carries `requirements.txt`, freeze hash matches source
- [x] Full project render (`quarto render .`) succeeds, 74/74; listing page and `search.json` pick the post up
- [x] `grep -rn "gsk_"` over `posts/`, `docs/`, `_freeze/`, `Makefile`, `CLAUDE.md` → clean; the key is read from the environment only
- [x] Prose word count 1751, under the 2000-word budget
- [x] Rendered page eyeballed in-browser: ToC reconstructs the argument, source blocks highlight, transcripts wrap instead of clipping


project-delphi (commented):
## Review — `feat(posts): LLM agents from first principles`

Scope reviewed: `posts/llm-agents-from-first-principles/{agent.py,index.qmd,requirements.txt}`, `Makefile`, `CLAUDE.md`. Generated output (`docs/`, `_freeze/`) checked for correctness and credential leakage but not line-reviewed.

### Critical Issues

None.

### Important Issues

**1. `calculator` was vulnerable to resource exhaustion (fixed in cab9114).** The AST whitelist correctly prevents arbitrary code execution — the stated goal — but `ast.Pow` was unbounded, so `9**9**9` parses to four perfectly legal nodes and hangs the interpreter for hours building an integer with billions of digits. This mattered more than usual here because the expression is *model-authored* (the post says so explicitly) and because the post advertises the AST walk as the safe alternative to `eval`. Confirmed by reproduction before fixing:

```
POW HANG CONFIRMED: timed out      # before
ERROR: calculator raised ValueError: exponent 387420489 exceeds the limit of 256   # after
```

Fixed with a `_MAX_EXPONENT` cap. Two things worth noting about the fix: the failure surfaces through `ToolRegistry.execute` as an observation string rather than an exception, which is exactly the design property the post argues for; and `_eval_node` is now displayed in the post, so the whitelist walk the prose describes is actually on the page instead of being asserted.

### Suggestions

- `GroqBackend.calls` is incremented but never read. It's self-documenting as an illustration of "how many times the policy was sampled", so this is deliberate rather than dead code, but a reader may look for the call site.
- `Agent.run` never passes `max_tokens`, so the `Backend` protocol's parameter is always defaulted to 512. Fine for these examples; a longer-horizon task would silently truncate.
- `is_prime` does `int(n)`, so a non-integral argument is truncated rather than rejected. The tool spec says `integer`, and the model has complied every run, but rejecting would be more honest than truncating.

### Non-findings (checked, no action needed)

- **Nested JSON in `ACTION_INPUT`.** `_INPUT_RE`'s non-greedy `\{.*?\}` looked like it would truncate `{"a": {"b": 1}}` at the first brace. It does not — the `\s*$` anchor forces backtracking to the full object. Verified: `parse(...)` returns `ToolCall(args={'a': {'b': 1}})`.
- **Credential handling.** The key is read from `os.environ` at call time, never a parameter, never a default, never written to disk. `grep -rn "gsk_"` over `posts/`, `docs/`, `_freeze/`, `Makefile`, `CLAUDE.md` is clean. The `HTTPError` path echoes the API's response body, not the request headers, so the key cannot leak through the error message.
- **Repo invariants.** `make check-posts` green; the post pins a dedicated kernel, carries `requirements.txt`, and its freeze hash matches source. `llm-agents` added to `kernels-stub` — without it a fresh clone could not render *any* page, since Quarto resolves kernelspecs at index time before consulting `_freeze/`. Full project render 74/74.
- **No fabricated output.** Both transcripts are cell output captured at render time. Arithmetic independently checked: `(37·4−9)/5 = 27.8`, `128²+3 = 16387`, `16387/7 = 2341` exactly, 2341 prime.

### Strengths

- The central design decision — `ToolRegistry.execute` converting every failure mode into an observation string rather than raising — is correct, deliberate, and explained where it happens rather than in a trailing caveats section.
- `parse()` returning `ParseError` as a *value* is the right call for a stochastic system, and the post earns the point by documenting a bug it actually hit (`ACTION: None` beside a valid `FINAL:`) instead of presenting a sanitized design.
- Zero third-party imports in `agent.py` makes the "first principles" claim literal rather than rhetorical, and reduces `requirements.txt` to Quarto's execution stack.
- `CLAUDE.md` correctly records the awkward part: this post's output is a *sample*, not a reproducible fixed point, so it must not be re-rendered to "refresh" anything.

**Verdict: no Critical issues. The one Important issue is fixed on this branch. Cleared to merge.**

PR #138, opened 2026-08-03
feat(posts): LLM agent memory — short-term vs long-term, via a voice assistant

project-delphi:
## What

A new ~1300-word conceptual post, `posts/llm-agent-memory/`, on how LLMs and agentic systems handle memory.

The spine: **an LLM is stateless per call**, so everything anyone calls "memory" is external engineering around that stateless core — what goes into the prompt, what is stored outside it, and when it is pulled back in.

One running example throughout, a TTS voice assistant taking calls from a user named Alex:

- **Short-term memory = the context window**, replayed in full every turn; the call is its lifetime. Truncation vs. compaction / context editing, and the fact that compaction is lossy on purpose.
- **System vs. user prompt as injection points, not memory** — slots where retrieved or summarised memory gets placed.
- **The four long-term options**, each judged against the voice app: RAG over a help-centre corpus, a key-value lookup for `user_id → voice_id` (on the TTS latency path, where a vector search does not belong), an agent-managed memory tool that decides for itself what is durable, and fine-tuning as the heavyweight wrong answer for personalisation.
- **Adjacent tooling**: compaction paired with a memory tool, prompt caching, vector DBs.
- **Two calls with Alex** — session one writes preferences out, session two assembles them back before the model is invoked, then falls back to RAG mid-call. Production layers several mechanisms rather than picking one.
- Closes on a single comparison table (concept / what it stores / where it lives / persists? / best for).

## Code

Four short cells, all stdlib-only, all executed at render and producing output: the transcript replayed on one turn, compaction of a long call, the agent's memory-tool write, and the session-two system prompt assembled from a KV lookup plus the memory file. No product names or invented numbers anywhere in the post.

## Environment

Not a legacy post. Pins the new `llm-agent-memory` kernel over `.venv-llm-agent-memory` — Quarto's execution stack only, the same shape as `.venv-llm-agents`, since the post imports nothing beyond the standard library — and carries `posts/llm-agent-memory/requirements.txt`. The kernel name is added to `make kernels-stub` so a fresh clone can render the site.

`docs/` and `_freeze/` are re-rendered in the same commit. `make check-posts` is clean; the full project render touched only `docs/index.html`, `docs/listings.json` and `docs/search.json` besides the new post.

## Preview

`python -m http.server 8000 --directory docs` → http://localhost:8000/posts/llm-agent-memory/index.html


project-delphi:
Added the Kokoro TTS demo (second commit on this branch).

**The voice ID is now real.** The key-value store returns `bm_george`, an actual [Kokoro-82M](https://huggingface.co/hexgrad/Kokoro-82M) voice pack, replacing the invented `calm-low-en-gb`. That closes the loop the post was gesturing at: the value long-term memory hands back *is* the file the speech engine loads.

**Two play buttons** sit after the "Two calls with Alex" walkthrough — the same assistant greeting the same caller, once with nothing stored (`af_heart`, the default voice) and once with the name read from the memory file and the voice from the KV lookup (`bm_george`). Same model, same code path; only what came out of storage differs.

**How the audio is built.** `src/make_audio.py`, run once with a borrowed `.venv-kokoro`, exactly the arrangement `src/make_cover.py` already uses in this post. The post's own kernel stays stdlib-only, so `requirements.txt` does not grow `torch` and no render downloads 327MB of weights. The script enumerates the voice packs published in the repo and fails loudly if one of the two is missing rather than trusting hard-coded names; resolved versions are frozen in `src/requirements-audio.txt`. Output is 64k mono MP3 resampled to 44.1kHz — 53KB for the pair, `resources: - audio` copies them into `docs/`.

**Verified:** both clips decode and play (`afplay` locally, and Chrome's WebAudio decoder in-page); the buttons play, pause, swap glyph, and stop each other; the clips serve 200 as `audio/mpeg`; one table still; `make check-posts` clean; body is 1400 words excluding code, at the agreed ceiling.

One caveat worth recording: I could not get audible playback through an `<audio>` element in the automation browser — but a synthetic WAV data URI stalls there identically, so it is that profile's media pipeline, not these assets.

project-delphi (commented):
## Review — PR #138

**Overview.** Adds `posts/llm-agent-memory/`: a 1400-word conceptual post (excluding code) arguing that an LLM is stateless per call, so everything called "memory" is external engineering, held together by one running example — a TTS voice assistant serving a caller named Alex. Four stdlib-only cells execute at render; a build-time Kokoro-82M script produces two demo clips played by buttons; one comparison table closes it. 17 files, source + `docs/` + `_freeze/` together.

### Critical Issues

None.

### Important Issues

Both found during this review and fixed in `3806e6b`'s follow-up commit, not deferred:

1. **`make_audio.py` documented how to *run* the build venv but not how to *build* it.** The recipe has a step nobody would reconstruct: misaki's English G2P loads a spaCy pipeline `kokoro` does not depend on, and `python -m spacy download` cannot install into a pip-less uv venv, so `en_core_web_sm` must come from its release URL. `requirements-audio.txt` pinned it, but only as one frozen line among 90. The build recipe is now in the module docstring.
2. **`SAMPLE_RATE = 24_000` was an unguarded, load-bearing constant.** Kokoro exposes its output rate nowhere — not on `KModel`, not on `KPipeline` (verified) — so it has to be hard-coded, and a wrong value produces mispitched audio rather than an error. That is a silent failure mode. It is now commented as such, with the instruction to re-check on a `kokoro` version bump. This is documentation rather than a code guard because the package gives nothing to assert against.

### Suggestions

- `synthesise()` flattens the tensor sample-by-sample in a Python comprehension. Fine for 4-second clips; `numpy.concatenate` is the idiomatic form if these ever grow.
- If `ffmpeg` fails, the intermediate `.wav` survives in `audio/`. Harmless as a build dir, but a `try/finally` would stop a stray WAV being swept into a later `git add -A`.
- The widget's `<style>`/`<script>` are inline in the `.qmd`; `posts/bayesian-bootstrap` externalises to `widgets.js`/`theme.scss`. At ~30 lines inline is defensible — worth hoisting if a second post grows an audio widget.
- `voices` in the post's cell is a dict literal standing in for a datastore. The comment says so, but a reader skimming could take it for a real client.

### Strengths

- **The environment separation is the right call and is applied consistently.** The post's kernel stays stdlib-only; Kokoro and its ~327MB of weights live in a borrowed venv exactly as `make_cover.py` uses `.venv-explainability`. `requirements.txt` does not grow `torch`, and no render downloads a model.
- **`make_audio.py` fails loudly rather than silently.** It enumerates the voice packs actually published in `hexgrad/Kokoro-82M` and exits with the missing names rather than trusting hard-coded strings, and raises if Kokoro yields no audio.
- **Deterministic output, verified.** Re-running the script produced different MP3 container bytes but bit-identical decoded PCM — the clips are reproducible, and the committed binaries were left untouched rather than churned.
- **The repo's invariants are respected and checked**: `make check-posts` clean, `_freeze` hash matches the source md5 exactly, `docs/` MP3s byte-identical to `posts/`, the new kernel added to `make kernels-stub` so a fresh clone can render, `LEGACY_NO_ENV` untouched.
- No secrets, no absolute paths, no stray build artifacts in the tracked file list.

### Test plan verified

- Both clips play locally (`afplay`) and decode correctly in Chrome's WebAudio decoder; buttons play, pause, swap glyph and stop each other; clips serve 200 as `audio/mpeg`.
- Full project render clean; exactly one `<table>`; all four cell outputs present; cover image resolves.
- Not verified: audible playback through an `<audio>` element in the automation browser — a synthetic WAV data URI stalls there identically, so it is that profile's media pipeline rather than these assets.

PR #139, opened 2026-08-03
feat(posts): a history of open-source TTS, with every era audible

project-delphi:
## Summary

A beginner-friendly history of open-source text-to-speech, built around one claim: **each era moved the unit of generation down a level of abstraction** — rules, then recorded fragments, then statistical parameters, then raw samples, then codec tokens — trading control for naturalness at every step.

- **The history is audible, not just described.** eSpeak NG and Flite still install, so four of the five demo clips are generated locally by `src/make_audio.py`, all speaking the same sentence so the only variable is technique: `espeak-ng` (formant), `flite cmu_us_kal16` (diphone), `flite cmu_us_slt` (Clustergen statistical parametric), and Kokoro-82M (codec era, via the existing `.venv-kokoro`). All normalised to −18 LUFS so clicking down the list compares technique rather than volume.
- **Interactive timeline** — vanilla HTML/CSS/JS, tablist semantics, arrow-key navigation. Selection is signalled by size, border weight and an underline *as well as* colour: browser auto dark mode flattens the purple to grey, which would otherwise leave no visible selection at all.
- **Comparison table** of five current models, with parameter counts and licences checked against primary sources.
- No executable cells, so the post pins no kernel and needs no `requirements.txt`; `src/requirements-audio.txt` freezes the build-time environment instead.

Two things worth a reviewer's attention:

- The **Tacotron 2 clip is the one we cannot generate** — the checkpoint was never released — so it is mirrored from `google.github.io/tacotron`. That repo carries **no explicit licence** (GitHub reports `NOASSERTION`). Mitigated by using one short clip, attributing it next to the player with a link back, and recording provenance in `make_audio.py`.
- **IndexTTS-2's parameter count is not published** in its repository, model card or paper, so the table says "not published" rather than guessing.

Prose is **1175 words** against the 1200-word brief (1269 if in-HTML audio captions are counted, which the brief exempts).

## Test plan

- [x] `make check-posts` passes
- [x] `QUARTO_PYTHON=... quarto render posts/open-source-tts-history/index.qmd` builds clean; `docs/` diff scoped to the new post plus `index.html`/`listings.json`/`search.json`
- [x] All 7 cited arXiv IDs fetched and confirmed to be the intended papers
- [x] Every external link returns HTTP 200 (festvox.org was dead, swapped to Flite's GitHub repo)
- [x] All 5 `<audio>` elements load in Chrome with correct durations, `readyState` 4, no media errors
- [x] Timeline: all 5 markers select exclusively via click and via arrow keys; panel content updates
- [x] No console errors; no horizontal page overflow; rail scrolls internally at 340px width
- [x] Loudness verified within 0.8 LU across all five clips (`ebur128`)
- [ ] **Human check needed:** listen to the five clips — audio quality is the one thing no automated check confirms


project-delphi (commented):
## Review: PR #139 — open-source TTS history post

**Overview.** Adds `posts/open-source-tts-history/` — a 1175-word history post with an interactive timeline, a 5-model comparison table, and five audio players, plus two build scripts and the re-rendered `docs/`. Source diff is ~500 lines across 4 new text files; the rest is the committed render output and 6 binaries (228 KB of MP3/PNG, reasonable for the content).

### Strengths

- **Follows the repo's documented conventions closely.** Build scripts live in `src/` and are explicitly *not* part of the post's execution environment, mirroring `llm-agent-memory/src/make_audio.py`. The module docstring records the exact rebuild commands, including the non-obvious `en_core_web_sm` wheel install that `python -m spacy download` can't do in a pip-less `uv` venv.
- **`SAMPLE_RATE = 24_000` is carried over with its rationale intact** — Kokoro exposes the rate nowhere, and a wrong value mispitches rather than errors. Correctly re-documented rather than silently copied.
- **No executable cells, so no kernel and no `requirements.txt`** — this is right, and matches `check_posts.py::check_environment`, which only requires an environment when a non-`eval: false` `{python}` cell exists. `make check-posts` passes.
- **Loudness normalisation is a substantive correctness choice, not decoration.** Without it the five engines differ by >10 LU and the comparison becomes one of volume; measured within 0.8 LU of −18 LUFS across all five.
- **Honest about what the artifacts are.** The prose states that eSpeak NG and Flite are living descendants rather than period binaries, and that Clustergen is the same family as HMM-based HTS but not literally HTS — a distinction easy to gloss over and verified here against `cmu_us_slt_cg` symbols in the Flite library.

### Important Issues

None that block merge.

### Suggestions

1. **`index.qmd` — timeline rail overshoots the outer dots.** `.ttshist-rail::before` is inset `left: 6%; right: 6%`, but the five dot centres sit at exactly 10/30/50/70/90% (measured). The connecting line therefore protrudes ~4% past the first and last dots as a small stub. `left: 10%; right: 10%` terminates it on the dot centres.

2. **`index.qmd` — `innerHTML` is unnecessary in `select()`.** The `data-title`/`data-body` attributes contain `&middot;`, but HTML attribute values are entity-decoded at parse time, so `dataset.title` already holds a literal `·` (confirmed: `hasEncodedEntity: false`). `textContent` is equivalent here and removes an HTML-injection sink should this data ever become dynamic.

3. **`index.qmd` — tab pattern is missing `aria-labelledby`.** The tabs carry `role="tab"` + `aria-controls`, but have no `id`s, and the panel has no `aria-labelledby` pointing back at the active tab. Giving each tab an `id` and setting `aria-labelledby` on the panel in `select()` would complete the pattern. Relatedly, `aria-live="polite"` on a `role="tabpanel"` can double-announce on some screen readers, since the tab pattern already announces the panel on selection — worth dropping one of the two.

4. **`make_audio.py` — no voice-pack validation.** The sibling script in `llm-agent-memory` validates voices against `list_repo_files("hexgrad/Kokoro-82M")` before synthesising and fails with a clear message. This one calls `KPipeline` directly, so a renamed/removed `af_heart` would surface as a less obvious runtime error.

5. **`make_audio.py` — the Homebrew engine versions aren't pinned anywhere machine-readable.** `requirements-audio.txt` freezes the Kokoro venv, but espeak-ng 1.52 and Flite 2.2 equally determine three of the five clips. They *are* recorded in the post's player labels, so this is a completeness nit rather than a reproducibility hole — a comment naming the tested versions in the docstring would close it.

6. **`Clip.detail` is overloaded** — it holds a Flite voice name, a Kokoro voice pack, or a URL depending on `engine`. Documented in the docstring and fine at 5 clips, but it's the field that would get awkward first if this grew.

### Correctness / security / performance

- No secrets in the diff. (Credential-shaped grep hits in `docs/search.json` are pre-existing published post content from the tokenizers and Snowflake posts, including a placeholder `your_password` — not introduced here.)
- `urlretrieve` targets a pinned `https` URL and raises on HTTP error, so a dead link fails the build rather than writing a partial file.
- `docs/` diff is correctly scoped: the new post plus `index.html`/`listings.json`/`search.json`. No collateral churn across other posts, and no `_freeze/` record is created or invalidated.
- 5 × `preload="metadata"` requests on load, ~146 KB of audio total — negligible.

### Verdict

**0 Critical, 0 Important, 6 Suggestions.** The two mechanical ones (rail geometry, `textContent`) are worth taking before merge since both are confirmed and one-line; the rest are optional polish.

One item genuinely needs a human and cannot be automated: **listening to the five clips**. Every other claim in the test plan was machine-verified.

project-delphi:
Applied in 6b85833 — suggestions 1, 2, 3 and 5 from the review above:

- Rail inset 6% → 10%, so the connecting line terminates on the outer dot centres instead of protruding past them (verified: inset now 10%, dot centres at 10/30/50/70/90%).
- `select()` switched to `textContent`; the only remaining mention of `innerHTML` is the comment explaining why it isn't used.
- Tabs gained `id`s, the panel gained `aria-labelledby` tracking the selected tab, and `aria-live` came off to avoid double-announcement. Re-exercised all five tabs: `aria-labelledby` follows selection and exactly one tab is selected at each step.
- Panel is now seeded with the first era's content in markup, so a reader without JS sees prose rather than an empty box.
- `make_audio.py` records the espeak-ng 1.52.0 / flite 2.2 versions the clips were built with.

Left as-is: suggestion 4 (Kokoro voice-pack validation — would add a network call to the build for marginal benefit) and 6 (`Clip.detail` overloading — documented, fine at five clips).

`make check-posts` passes and the post re-renders clean; prose still 1175 words.

PR #140, opened 2026-08-03
New post: intelligence as generalization, and why the LLM stack rhymes with it

project-delphi:
## Summary

A short prose post (~550 words) at `posts/intelligence-for-novel-problems/`, arguing one thing across three sections:

- **Skill can be bought; generalization has to be earned.** What's worth measuring is the exchange rate from experience to competence on problems outside the training distribution — Chollet's framing, linked inline. Memorization buys the first and not the second.
- **An unstable niche made the general-purpose engine worth its price.** Instinct is cheap and brittle; a general inference engine is metabolically ruinous but adapts within a lifetime. Tool making is that engine's visible output *and* a feedback driver — calories funding brain tissue, dexterity and planning coevolving, cumulative culture ratcheting language and social cognition along.
- **No single paradigm reproduces it; the stack roughly does.** Self-supervised ≈ childhood world-modelling, supervised ≈ cultural transmission, RL ≈ consequence (and evolution at species scale). None alone gives few-shot competence; the integration does, which is why pretraining → SFT → RLHF is no coincidence. Closes on where the analogy stops holding — a child intervenes on the world rather than reading descriptions of it.

Prose-only, so no `jupyter:` kernel, no `requirements.txt` and no `_freeze/` record — `check_posts.py` skips both of its checks for a post with no executable cells. `src/make_cover.py` follows the existing house-style generator in `posts/open-source-tts-history/src/make_cover.py` (solid `#4A3AA7`, badge pill, triple-ring logomark) and is run with any venv that already has matplotlib rather than adding a fourteenth environment for one PNG.

Source and re-rendered `docs/` land in the same commit, per CLAUDE.md.

## Test plan

- [x] `make check-posts` → `OK (5 legacy posts grandfathered)`
- [x] `QUARTO_PYTHON="$(pwd)/.venv/bin/python" quarto render .` → 77/77 pages, `Output created: docs/index.html`; no post re-executed (`freeze: auto`)
- [x] Render diff is confined to the new post plus `docs/index.html`, `docs/listings.json`, `docs/search.json` — no unrelated churn
- [x] `cover.png` is 1200×630 and visually matches its siblings
- [x] Body word count 549, inside the 450–550 target


project-delphi (commented):
## Review

**Overview** — Adds one prose-only post (`posts/intelligence-for-novel-problems/`, 549 body words) plus its house-style cover generator, and commits the re-rendered `docs/` in the same commit as required by CLAUDE.md. No executable cells, no kernel, no `requirements.txt`, no `_freeze/` record.

### Critical Issues

None.

### Important Issues

None.

### Suggestions

- **`src/make_cover.py` trips ruff `D406`/`D407`** (`Returns:` wants a dashed underline under the repo's pydocstyle setting). Not a regression — `posts/open-source-tts-history/src/make_cover.py` and `posts/llm-agent-memory/src/make_cover.py` trip the identical two rules, so the Google-style docstring here matches its siblings rather than the configured convention. Worth fixing all three together at some point, not in this PR.
- **`mypy` reports `import-not-found` for `matplotlib`** on the new file. Expected: matplotlib isn't in the base `.venv` (`uv sync` prunes to the lock), and the same is true of every sibling generator. Not actionable.
- **The expensive-tissue hypothesis is alluded to but not named** ("Better calories helped fund the tissue doing the thinking"). Naming it, or linking Aiello & Wheeler, would give a reader a search term — at the cost of words in a post already at the top of its length target.

### Verification performed

- `make check-posts` → `OK (5 legacy posts grandfathered)`. Confirmed the checker genuinely covers this post: `check_environment` is skipped because no non-`eval: false` `{python}` cell exists, and `check_freeze` passes because no freeze record exists and the slug isn't in `LEGACY_NO_ENV`. Nothing to register.
- **`docs/search.json` churn is additive, not a rewrite.** The 1140-line diff is reindexing noise; parsed both sides, 549 → 553 entries, the 4 added are all the new post's page and its three section anchors, **zero removed**, and exactly one shared entry changed content (`index.html`, the listing page). No existing post lost its index entry.
- **`docs/index.html`** gained the listing row (`listing-image` + `listing-title`) pointing at the new cover and page.
- **Rendered output is correct**: `<title>` set, three `<h2>` anchors matching the source headings, ToC present, `src="./cover.png"` resolves, arxiv link intact at `1911.01547` (correct ID for *On the Measure of Intelligence*), and zero unrendered markdown constructs left in the HTML.
- `cover.png` is 1200×630 RGBA, solid `#4A3AA7` with badge pill and triple-ring logomark — matches its siblings. `black --check` passes on the generator.
- Render was `QUARTO_PYTHON`-pinned and completed 77/77; `freeze: auto` held, so no existing post re-executed and no unrelated `docs/` churn appears in the diff.

### Strengths

- Frontmatter matches the modern template (`posts/nvidia-buys-the-pyg-team`) field for field, including the optional `description`, and correctly **omits** `jupyter:` — which is what keeps this post out of the environment check.
- Headings state claims rather than topics, so the ToC reconstructs the argument, per the writing conventions.
- The caveat ("a child's world model is built by intervening, not by reading descriptions") sits inline at the point the objection arises rather than in a Limitations bin at the end.
- `src/make_cover.py` reuses the existing generator rather than inventing a new cover pipeline, and deliberately borrows an existing venv instead of adding a fourteenth environment for one PNG.
- British spelling ("behaviour", "modelling") is consistent with the rest of the blog.

PR #142, opened 2026-08-06
feat(posts): can you invert a recursive function

project-delphi:
## Summary
- New short post distinguishing recursion (a definitional/implementation shape) from invertibility (an injectivity property).
- `to_binary`/`from_binary` as an information-preserving recursive bijection; `total(xs)` as a lossy accumulator where distinct lists collapse to the same sum.
- Closes with a speculative analogy: memory/perception as a similarly lossy, non-injective encoder — a loose thematic companion to the recently-merged `consciousness-recursive-prediction` post.
- Kernel: `blog-base` (stdlib only), with its own `cover.png`, `requirements.txt`, and `_freeze/` record per the repo's post conventions.

## Test plan
- [x] `make check-posts` passes.
- [x] `QUARTO_PYTHON=.venv/bin/python quarto render .` — full site render succeeds, both code cells execute and render live output (`[1, 0, 1, 1]`, `True`, `6`).
- [x] Verified rendered HTML locally via `python3 -m http.server --directory docs`.


project-delphi:
## Review (via /code-review)

Two findings, both addressed in 7330142:

1. **`requirements.txt` was not the post's actual runtime.** It was a `pip freeze` of the author's base dev-tooling venv, including a machine-local `-e file:///Users/ravikalia/Code/github.com/ml-blog` editable-install line that would break `pip install -r` on any other clone.
2. **`jupyter: blog-base` contradicted CLAUDE.md's documented scope for that kernel** — blog-base is for `eval: false` display-only posts, but this post's cells genuinely execute.

**Fix:** gave the post its own `.venv-recursive-inversion` + named `recursive-inversion` kernel per the standard per-post recipe, re-rendered, and regenerated `requirements.txt` from that dedicated venv (now just Quarto's execution stack — ipykernel/jupyter/nbclient/nbformat/pyyaml — since the post itself imports nothing but stdlib). `make check-posts` passes.

PR #143, opened 2026-08-06
feat(posts): structured JSON output from the Claude API

project-delphi:
## Summary

New post, `posts/structured-json-with-claude/`, on getting schema-conforming JSON out of the Claude API instead of parsing prose.

- Frames the argument around the interface mismatch: a model tuned to be conversational will wrap a payload in preambles and fences, and fuzzy text/data boundaries are exactly what parsers can't tolerate.
- Contrasts the **old** approach (prefill the assistant turn with `{`, clip the reply with `stop_sequences`, glue the braces back on) against the **new** one (`output_config.format` with a JSON Schema, enforced at the endpoint).
- Notes plainly that prefill now returns a 400 on current models, so the old pattern is a time bomb rather than merely dated.
- Closes on the generalization — anything where model output becomes a variable rather than a paragraph — plus two honest caveats: a schema guarantees shape, not correctness, and `stop_reason` can still signal a refusal or a `max_tokens` truncation that won't conform.

Two details worth flagging for review, both corrected against the current API docs while drafting:

- The model is `claude-sonnet-5`, not `claude-sonnet-4-5`. The latter isn't on the structured-outputs supported list, so the example would not have worked as written.
- The schema carries `additionalProperties: false`, which structured outputs requires on every object.

Cover follows the house style (`#4A3AA7` field, category pill, triple-ring logomark) via `src/make_cover.py`, matching the generator pattern used by recent posts.

## Test plan

- [x] `make check-posts` passes — post is display-only (all cells `#| eval: false`, pinned to the shared `blog-base` kernel), so no dedicated venv or `requirements.txt` is required, same as `posts/langgraph-vs-llamaindex`.
- [x] Full site render (`QUARTO_PYTHON=... quarto render .`) completes; `freeze: auto` held, so no other post's output moved — the only tracked changes are `docs/index.html`, `docs/listings.json`, `docs/search.json`.
- [x] `search.json` verified programmatically: 567 → 574 entries, nothing removed, the only changed pre-existing entry is the site index page that now lists this post. The large line diff is reordering churn, not content loss.
- [x] Both `{python}` blocks parse via `ast.parse`; the sample response block parses via `json.loads`.
- [x] Rendered page eyeballed in `quarto preview`, cover image included.
- [x] No live API call is made at render time and no key is required; the sample response is hand-written and labelled as such in the post body.
- [ ] Not verified: the code example has not been executed against the live API, by design — the post states this explicitly rather than presenting a captured transcript.


project-delphi (commented):
## Automated review (`/code-review`)

**Verdict: 0 Critical, 0 Important, 4 low-severity findings.**

### Mechanical verification (all pass)

- `python3 scripts/check_posts.py` → `OK (5 legacy posts grandfathered)`. The `_freeze` hash matches the source; both `{python}` cells are `#| eval: false`, so the post correctly needs no dedicated venv/`requirements.txt` and correctly pins `jupyter: blog-base`.
- `docs/` re-rendered in the same commit (per CLAUDE.md). `docs/search.json` grew 567 → 574 entries with **zero** existing `href`s dropped — the 4035-line diff is reordering, not loss. `docs/index.html` only shifts `data-index` values.
- `posts/.../cover.png` and `docs/posts/.../cover.png` are the identical blob (`d1b23d4`). `src/` is not copied into `docs/`, consistent with the seven other posts that have a `src/`.
- Rendered HTML resolves the cover, the JSON block, and both reference links.

### Findings

**1. `index.qmd:54` — "returns a 400 on current Claude models" is overbroad.**
The enumeration that follows ("Sonnet 5, Opus 5, and the 4.6-and-later family") is correct, but the bolded blanket claim is not: Haiku 4.5 is a current, non-retired model and still accepts assistant-turn prefill. A reader running prefill on Haiku 4.5 reads "it isn't legacy-but-working. It's a time bomb" and rips out code that works today.

**2. `index.qmd:90` — `max_tokens=1024` can trigger the exact failure the post promises won't happen.**
Sonnet 5 runs adaptive thinking when `thinking` is omitted (unlike Sonnet 4.6, which ran thinking-off by omission), and `max_tokens` caps thinking *plus* response text together at default `effort: high`. On a longer ticket than the toy one shown, thinking can eat most of the budget and the object truncates mid-emit — `stop_reason: "max_tokens"`, non-conforming output — immediately after the post asserts "Just an object matching the schema, every time."

**3. `index.qmd:95` — the sample doesn't do what the post's own caveat tells readers to do.**
`text = next(b.text for b in response.content if b.type == "text")` — line 117 correctly says to check `stop_reason` before parsing, but the copy-pasteable block doesn't. On a pre-output refusal `content` is empty, so this raises a bare `StopIteration` rather than a diagnosable error.

**4. `index.qmd:60-73` — synthetic input isn't situated per CLAUDE.md.**
The repo's "Always situate the data" convention explicitly covers synthetic data. The post discloses that the *response* is hand-written but introduces the ticket as a bare "Here's a messy support ticket." Borderline — a single fabricated input string is arguably not "a dataset" — but flagged since the convention is stated in absolute terms.

### Reviewed and clean

- The "old way" block is technically accurate as depicted: prefill text is not echoed in the response, and a matched `stop_sequence` is excluded from output, so both the `"{" +` and `+ "}"` re-attachments are correct. The nested-object critique of `stop_sequences=["}"]` is also correct.
- `output_config={"format": {"type": "json_schema", "schema": schema}}` is the current (non-deprecated) parameter shape on stable `messages.create`, and Sonnet 5 supports structured outputs. The schema satisfies strict-schema requirements (`additionalProperties: False`, all properties in `required`, only supported keywords).
- `src/make_cover.py` is sound: `matplotlib.use("Agg")` precedes the `pyplot` import, `OUT` resolves to the post directory, pill and logomark don't overlap, `SystemExit(main())` propagates correctly. Only nit is a 2px vertical miscentering of the badge text — not worth changing.

project-delphi:
Addressed all four findings in `429e4c5` (none were merge-blocking, but three were genuine accuracy problems worth fixing before publishing):

**1. Overbroad prefill claim — fixed.** The blanket "returns a 400 on current Claude models" now reads: *"prefill now returns a 400 on Sonnet 5, Opus 5, Fable 5, and the 4.6-and-later Opus and Sonnet models. Haiku 4.5 still accepts it, so it isn't dead everywhere — but 'works on the model I happen to call' is a thin foundation to build on."* Keeps the "don't build on this" advice without telling a Haiku 4.5 user their working code is broken.

**2. `max_tokens` vs. adaptive thinking — fixed.** Raised to 4096 with an inline comment (`# shared by thinking + response; leave headroom`), and the closing caveat now states that thinking shares the budget. The overclaim that sat right after the code ("Just an object matching the schema, every time") is now scoped to a turn that completes.

**3. Sample contradicting its own caveat — fixed.** The block now guards before parsing:
```python
if response.stop_reason != "end_turn":         # refusal, or truncated mid-object
    raise RuntimeError(f"no conforming output: {response.stop_reason}")
```
The closing paragraph points at it explicitly rather than describing a check the code didn't do.

**4. Unsituated synthetic input — addressed.** The ticket is now introduced as *"invented for this post, but written the way an annoyed customer actually writes."* Light touch, since a single fabricated string isn't really a dataset, but it satisfies the convention.

The badge-centering nit was left alone as suggested.

Re-verified after the edits: `make check-posts` initially **failed** on a stale freeze hash (working exactly as intended — the edits invalidated the record), so the post was re-rendered and now passes. Blast radius is still tight: `docs/index.html` 1 line, `docs/search.json` 3 lines, plus this post's own output and freeze record. No other post moved. Prose is 771 words.

PR #146, opened 2026-08-08
feat(posts): the Messages API's two delivery modes, and where they diverge

project-delphi:
Adds **"Waiting, and Not Waiting"** (`posts/messages-api-streaming/`) — an introduction to the Claude Messages API for a reader who writes code competently but has never made an LLM API call.

## Why this post

The blog's only existing Claude API post, `structured-json-with-claude`, answers a narrow question: how to get JSON out instead of prose. Nothing on the blog explains the endpoint itself. This one sits underneath it.

It has a single spine, and every section advances it:

> A full response and a stream are the same request delivered two ways — so the choice is a UX decision, right up until tool calls turn it into a correctness one.

That framing is what lets the partial-JSON gotcha land as the payoff of the argument rather than as a bullet in a list of caveats.

## What it covers

One endpoint and its statelessness → the non-streaming call and why it's the right default → SSE from first principles, with the two real motivations (perceived latency, and timeouts on long generations) → the event sequence as nested envelopes → `input_json_delta` fragments → a decision guide → error handling and non-resumability.

## Accuracy

Every API claim is checked against the current streaming documentation rather than recalled. The two details I could not source offline were verified before writing:

- Tool arguments stream as `input_json_delta` events carrying a `partial_json` **string**, while `tool_use.input` is *always* an object in the final message — so fragments must be concatenated and parsed once, at `content_block_stop`.
- Tool use and thinking blocks "cannot be partially recovered"; you can only resume from the most recent text block. The post explains *why* in terms it has already established — broken JSON is not a value, and a thinking block is closed by a `signature_delta` that an unsigned fragment can't reproduce.

No invented behaviour, and no TODOs left in the text.

## Repo mechanics

- **Display-only post.** Every `{python}` cell carries `#| eval: false` and it pins the shared `blog-base` kernel, so it owns no venv and no `requirements.txt`. `make check-posts` passes.
- **Cover** generated by `src/make_cover.py` in the house style (`#4A3AA7`, category pill, triple-ring logomark), run ephemerally via `uv run --with matplotlib` so nothing lands in `.venv`.
- **Source and rendered `docs/` committed together**, as required — the diff to `docs/` is this post plus `index.html`, `listings.json` and `search.json`.
- A pre-existing fenced-div warning appears during a full-site render; rendering this post in isolation produces no warning, so it is not from this change.

## Note for review

The brief asked for ~1200 words of prose; this lands at roughly 1400 including headings, after a trimming pass. Happy to cut a further section if that's over budget — the "Which one to reach for" table and the closing exercise are the most compressible parts.


project-delphi (commented):
## Automated review — `code-review` (0 Critical, 3 Medium, 3 Low)

Merge gate: **no Critical issues**. All six findings were nonetheless fixed in 55135ee before merge, because three were factual errors about the API in a post whose stated premise is that its claims are verified.

Root cause of the three Mediums was the same: the post was written as though `claude-opus-5` behaved like its predecessors. It doesn't — **thinking is on by default** when the `thinking` parameter is omitted (a change from Opus 4.8/4.7), so a `thinking` block appears in `content`, and `max_tokens` caps thinking and text together.

### Medium

1. **`index.qmd:33` — `content` enumeration was wrong.** Described as "usually one text block, sometimes text plus a tool call". On Opus 5 a `thinking` block also arrives ahead of the text. A reader following the post's own "check `block.type`" advice would have hit a type the post never listed, in the one section whose job is to enumerate them. *Fixed: the thinking block is now named, and the advice is stated as "check `block.type` rather than reaching for `content[0].text`".*

2. **`index.qmd:47` — first sample would truncate as a template.** `max_tokens=1024`, no `thinking` field. Fine for the paperclip prompt, but lifted as a starting point with a real prompt it burns the budget before the answer starts: `stop_reason: "max_tokens"`, and the sample's `if block.type == "text"` loop prints a fragment or nothing with no visible cause. *Fixed: raised to 4096, with an inline comment and a paragraph explaining that the cap covers thinking too.*

3. **`index.qmd:171` — the exercise promised event types it wouldn't produce.** It called `for event in stream:` "the raw event loop" and gave an exact expected list. `messages.stream()` is the accumulating helper. *Fixed: switched to the genuinely raw iterator, `client.messages.create(..., stream=True)`, with a runnable snippet. The expected list now accounts for the thinking block — opens, takes one `signature_delta`, closes — which turns the exercise into a payoff for the non-resumability section rather than a contradiction of it.*

### Low

4. **Retry framing.** "the SDK already retries it twice by default" sat next to the mid-stream `error` event, implying `max_retries` covers the failure the post recommends streaming for. It doesn't — the SDK cannot rewind a stream. *Fixed: the callout now says what it doesn't cover.*
5. **"one more we're about to get to"** contradicted the post, which introduces `signature_delta` later. *Fixed: listed, and the enumeration is now explicitly open rather than exhaustive.*
6. **Dangling pronoun.** "It cannot show you anything until it can show you everything" opened the streaming section referring to the *previous* section's subject, so it read as the inverse of the claim. *Fixed.*

### Verified clean by the review

`_freeze` hash matches `md5(index.qmd)`; `make check-posts` passes; post is correctly display-only on `blog-base` with no venv; `docs/search.json` gained 10 entries and lost 0; no other post's rendered output changed; cover PNGs byte-identical between source and `docs/`.

### Out of scope, worth a separate issue

`_freeze` records for `posts/pinecone-vs-weaviate` and `posts/pyspark` don't match their sources' md5 — stale caches a project render would try to re-execute. Untouched by this PR, but they're on `check_posts.py`'s `STALE_FREEZE_OK` list precisely so this stays invisible.

### Note on length

The brief asked for ~1200 words of prose. After the accuracy fixes this is ~1470 excluding headings. I couldn't get to 1200 without dropping required coverage — the most compressible parts remain the "Which one to reach for" table and the closing exercise.

PR #147, opened 2026-08-09
feat(posts): LLM API pricing compared, August 2026 snapshot

project-delphi:
## Summary

- New post `posts/llm-api-pricing-2026/` comparing per-token API pricing across seven providers (Anthropic, OpenAI, Google, DeepSeek, Moonshot/Kimi, Alibaba/Qwen, xAI) as of 8 August 2026.
- Framed explicitly as a **dated snapshot**, not a standing reference — several providers repriced multiple times recently, and Sonnet 5's intro rate expires 31 Aug 2026. The closing section tells readers to verify against the official pricing pages before budgeting.
- The argument isn't "here are some numbers": it's that three structural differences move a real bill more than the headline rate does — long-context surcharges (OpenAI past 272K, Google past 200K, Anthropic none), caching discounts (75–98% off cached input), and which providers treat audio/video as first-class inputs.
- Per-provider tables plus one combined 24-row table sorted by input price, with numeric columns right-aligned. Plain Quarto markdown — no OJS/DT widget, so no runtime dependency for a table that never changes after publish.
- Cover generated by `src/make_cover.py` in the house style (#4A3AA7 field, category pill, triple-ring logomark), matching `posts/messages-api-streaming`.

## Notes for review

- **No code cells**, so the post pins no `jupyter:` kernel and needs no `.venv-<slug>` or `requirements.txt`. `scripts/check_posts.py` treats it as a prose post; it does not enter `LEGACY_NO_ENV`.
- Prices were supplied as source data for the post rather than researched during authoring, so the tables are a faithful transcription of the provided figures, not independently verified against provider pages. The reference list carries the primary sources.

## Test plan

- [x] `make check-posts` — OK (5 legacy posts grandfathered), new post raises nothing
- [x] Full site render: `QUARTO_PYTHON="$(pwd)/.venv/bin/python" quarto render .` — 82/82 documents, exit clean
- [x] Render diff is scoped: only the new post's output plus `docs/index.html`, `docs/listings.json`, `docs/search.json` — no re-execution churn in unrelated posts (freeze respected)
- [x] Post appears in the site listing and in `search.json`
- [x] Rendered HTML has 8 tables with numeric columns right-aligned
- [ ] Not verified: the pricing figures themselves against live provider pages


project-delphi (commented):
## Code review — `/code-review high`

Reviewed the PR diff (one commit, clean working tree). Anthropic pricing and context claims were verified against the authoritative `claude-api` skill reference rather than from memory. `make_cover.py` has no correctness defects (data units map 1:1 to pixels at 1200×630/100dpi, so the `Circle` patches render circular), and the `$/M` table headers don't trigger pandoc math mode — the closing `$` is space-preceded, confirmed in the rendered HTML.

Every finding is in the post's prose/tables and is mirrored into the committed `docs/` output, so each fix needs a re-render in the same commit per the repo's `docs/`-in-sync rule.

### High

**`index.qmd:28` — Haiku 4.5 context window is 200K, not 1M.**
The Anthropic table lists `| Haiku 4.5 | 1.00 | 5.00 | 1M |`. Anthropic's published context window for `claude-haiku-4-5` is **200K** — it is the one current model that is not 1M. Repeated at `index.qmd:135` in the combined table, and published at `docs/posts/llm-api-pricing-2026/index.html:216` and `:619`. A reader budgeting a long-context workload picks Haiku 4.5 at $1.00/M expecting 1M-token prompts and gets a 400 at ~200K. Also invalidates the section's framing that Anthropic's ladder is uniformly 1M.

**`index.qmd:152` — "DeepSeek V4-Flash is the overall price floor" contradicts the post's own table.**
The combined table is sorted by input price and puts **Qwen3.5 Flash at $0.10** ahead of V4-Flash at $0.14 (`:126-127`), and the Qwen section at `:100` annotates Qwen3.5 Flash as "Cheapest rate in this comparison". V4-Flash is only the floor on *output* ($0.28 vs $0.40). Same claim again at `:69`.

### Medium

**`index.qmd:69` — "undercuts every Western tier by roughly an order of magnitude" is false against the post's own numbers.**
V4-Flash $0.14 vs GPT-5.6 Luna $0.20 is 1.4×; vs Grok 4.1 Fast $0.20, 1.4×; vs Gemini 3.5 Flash-Lite $0.30, 2.1×. The order-of-magnitude gap holds only against Western *flagships*.

**`index.qmd:114` — "Largest context window here" is a tie, not a distinction.**
Grok 4.1 Fast is annotated as largest at 2M, but Gemini 3.1 Pro is also 2M (`:60-61`, `:140`, `:144`).

**`index.qmd:110` and `:156` — Grok's "unusually low ~3× output:input ratio" is contradicted by DeepSeek in the same table.**
Both DeepSeek models are exactly **2×** (0.14→0.28, 0.435→0.87), lower than every Grok model. Grok's own three are 2.5×, 2.0×, 3.0×, so "roughly 3×" is the top of its range. "5–6× is the norm elsewhere" also fails against Qwen3.8-Max (3×) and Qwen3.5 Flash (4×).

**`index.qmd:154` — "Claude uniquely has no long-context surcharge" overreaches.**
The post documents surcharges only for OpenAI (272K) and Google (200K); DeepSeek, Moonshot, Qwen and xAI are all listed at 1M–2M with none. Anthropic is unique *among the Western majors* — which is how the Anthropic section correctly phrases it at `:24` ("the only major provider"). The takeaway drops the qualifier and becomes false.

### Low

**`index.qmd:31` — Mythos 5 listed as a purchasable tier with no availability caveat.**
`| Fable 5 / Mythos 5 | 10.00 | 50.00 | 1M |` presents the two as interchangeable SKUs. Claude Mythos 5 is access-restricted (Project Glasswing participants only), not generally available at that or any price. Same at `:147`.

**`index.qmd:24` — the "4.6 and newer" caveat excludes the first row of its own table.**
"the rate is flat regardless of prompt length on models 4.6 and newer" is immediately followed by a table whose first entry is Haiku **4.5**.

**`index.qmd:35` and `:155` — caching discount stated without the cache-write premium.**
"~90% off cached input" is correct for cache *reads* (~0.1× base input), but cache *writes* cost **more** than base input — 1.25× at the 5-minute TTL and 2× at the 1-hour TTL. Break-even is two requests at 5m, three at 1h. Since `:155` claims caching "dominates real-world cost", a reader modelling a low-reuse workload will project savings where caching is a net cost.

**`index.qmd:148` — audio-token price sorted into a text-token price column.**
GPT-Realtime-2.1's $32.00 is per M **audio** tokens (correctly noted at `:46`) but is sorted as the most expensive row of a table headed "Input $/M" whose other 22 rows are text tokens. The sort implies it is 3.2× the price of Opus 5 on a comparable unit.

### Strengths

Render scope is correct (only the new post plus `docs/index.html`, `listings.json`, `search.json`); the post indexes into `search.json`; `make_cover.py` is clean.

PR #148, opened 2026-08-09
feat(posts): cascaded vs native voice AI architectures, August 2026

project-delphi:
## Summary

A new post at `posts/voice-ai-architectures-2026/` on the voice AI stack: both architectures, four independently-run benchmarks, the vendor landscape, per-component picks, and a calibrated forward view.

- **The argument changed during research, for the better.** The stock framing — cascade is slower but more accurate on real audio — is contradicted by the evidence. Full-Duplex-Bench-v3 runs a cascade and GPT-Realtime on the *same* GPT-4o-class LLM: the cascade is slower **and** less accurate (0.450 vs 0.600 Pass@1), collapsing to 0.176 on self-corrections because Whisper finalises the transcript before the user's correction lands. The post therefore argues the transcript is a *lossy, early, irreversible commitment* — the property that makes it auditable is the one that loses the correction.
- **Sourcing is primary-only.** Aggregator posts dominate search for these terms and are wrong on live data (the widely-repeated "Cohere Transcribe tops Open ASR at 5.42%" is stale on both counts). Numbers come from arXiv papers, model cards, live leaderboard data and vendor pricing pages, each with a retrieval date.
- **Figures are reproducible without a venv.** `src/make_figs.py` holds every number in one dict with a primary source and a `self_reported` flag per row; PNGs are pre-rendered, so the post has no executing cells and stays out of the freeze regime (no kernel, no `requirements.txt`).
- Unverifiable claims are flagged inline rather than dropped silently — including a genuine discrepancy between PersonaPlex's model card and its paper.

## Test plan

- [x] `make check-posts` passes (post trips neither invariant)
- [x] Kernelspec pre-flight — all 16 pinned kernels registered before the full render
- [x] Full site render with `QUARTO_PYTHON` set; `docs/` diff scoped to the new post + `index.html`/`listings.json`/`search.json`, no unrelated post re-executed
- [x] All four figures + cover referenced and present in rendered HTML
- [x] Mermaid diagram renders in-browser, cascade left / native right in reading order
- [x] 8 footnotes resolve with back-links; no stray LaTeX math from `$` in prices
- [x] Chart palette validated for CVD and contrast (worst adjacent pair ΔE 27.1 protan)
- [x] Every figure eyeballed at full size for label collisions


project-delphi (commented):
## Automated review (`/code-review 148`)

**Verification performed:** all four figures regenerated from `make_figs.py` in a scratch dir and confirmed byte-identical to the committed PNGs; `posts/` and `docs/` image copies match; `docs/search.json` gained only the new post's 17 entries with zero removals (no other post disturbed); `scripts/check_posts.py` passes; rendered HTML matches source (prices render as `$…`, footnotes back-link, mermaid emitted as `pre.mermaid`, all 5 images resolve). Ruff/black failures in the two new scripts match the pre-existing baseline for every other `posts/*/src/*.py` and were not treated as findings.

### Findings, most important first

1. **`index.qmd:328` — prose contradicts the adjacent table.** "150× price difference from ElevenLabs Multilingual v2 for a gap of about 40 Elo points" cites a model in neither the table nor `make_figs.py`. The price ratio is Eleven **v3** ($100.00/$0.65 = 154×), but that pair's Elo gap is 1171−1056 = **115**, not 40 → ~66/34 preference, not "roughly 56/44". Same mismatched pairing is the premise of a forecast at line 504.
2. **`index.qmd:466` — "a third to a fifth of the ElevenLabs price" is wrong for one of two models named.** Cartesia Sonic 3.5 is $49.00 vs $100.00 — a *half*. Only Gemini 3.1 Flash TTS ($18.30) fits.
3. **`index.qmd:310` — TTS table shows 11 rows; the figure above plots 13.** `Mistral Voxtral TTS` (1067) and `NVIDIA Magpie-Multilingual 357M` (1065) are plotted but absent from the table, and from the vendor table.
4. **`make_figs.py:306` — label collision in `fig-s2s-gap.png`.** The `nudge` dict separates GPT-Realtime from Grok but not Qwen Audio 3.0 Realtime Plus (99, 98.4), whose left-flowing label overruns the GPT label nudged up from (96, 95.7).
5. **`index.qmd:116` — citation internally inconsistent.** PersonaPlex footnote gives `arXiv:2602.06053` but says "submitted 14 January 2026"; arXiv IDs encode YYMM, so January would be `2601.*`.
6. **`index.qmd:352` — S2S table is not in the descending order it appears to use.** 99, 97, 96, 97 — Gemini at 97 sits below GPT at 96, and it isn't a dynamics tiebreak. `make_figs.py` orders the same rows differently again.
7. **`index.qmd:268` — "seven English short-form corpora" vs six named.** Reconciles only if LibriSpeech clean/other count separately, which the post never says.
8. **`index.qmd:310` — the price column mixes two incompatible kinds of number.** Vendor list rate for proprietary rows; some third party's hosting rate for open-weights rows (Kokoro $0.65, XTTS v2 $40.40 — the latter for a project whose company shut down). No host given, yet the 150× headline and the cost forecast rest on it.
9. **`index.qmd:422` — vendor table omits three vendors the post benchmarks by name.** xAI (Grok Voice, in two charts), Fixie (Ultravox v0.7, leads the latency chart) and Deepslate (Opal) have no row, under a "every row was checked" framing followed by a no-consolidation completeness claim.
10. **`index.qmd:145` — the RTF/RTFx glossary entry defines only RTFx.** Applying the given definition to RTF inverts it (RTF > 1 means *slower* than real time).
11. **`index.qmd:4` — frontmatter `description` understates and partly contradicts the body.** It says native models "closed the latency gap"; the body's finding is they are *ahead* by ~2.4×, and the post never measures the two architectures' output audio quality.
12. **`make_figs.py:164` — minor: legend order runs opposite to visual order** in `fig-latency.png`; `invert_yaxis()` puts the task-completion bar on top within each pair.

project-delphi (commented):
## Automated review, round 2 (`/code-review 148`)

No Critical issues. 13 findings (6 Medium, 7 Low), all data/prose-consistency rather than correctness-of-code. **All addressed in 572336d except one, which I checked and rejected — details below.**

### Medium

1. **`index.qmd:229`** — the "2.4×" latency win averages Gemini Live 3.1 over only the 78% of runs where it spoke, against a cascade with a 100% turn-take rate. Survivorship-biased comparison, and it is the number actually quoted. → caveat added as the first of three.
2. **`index.qmd:500`** — "GPT-Realtime-2.1 … leads Full-Duplex-Bench-v3 on task completion" contradicts the post's own table, whose *Task completion (s)* column places it 4th of 6. Intended claim was Pass@1 success. Also, FDB-v3's footnote points at `gpt-realtime-1.5`, so it is not evidence about 2.1. → both corrected.
3. **`index.qmd:507`** — "reasons better than the pure speech-token models" reads as a claim Nemotron tops open reasoning; Freeze-Omni scores 33 vs 27. → scoped to the speech-token models by name, with Freeze-Omni stated explicitly.
4. **`make_figs.py:356`** — chart title generalised PersonaPlex's dynamics score to all five open models; four sit below every proprietary model on that axis. → retitled.
5. **`index.qmd:331`** — the † price note described four of the five rows it covered incorrectly (Fish Audio and Voxtral prices *are* the authors' own API rates; Coqui's is not). → rewritten, markers moved.
6. **`index.qmd:171`** — claimed TEDLIUM is missing from the enumerated corpora, making it eight test sets. **Rejected.** Reproduced the published `avg` from the per-dataset columns for three models (vfast 4.4300 vs 4.43; canary-qwen 5.0643 vs 5.06; granite-4.1 4.9014 vs 4.90) — the column used averages exactly the seven listed sets. Wording tightened to state this and how it was verified.

### Low

7. `index.qmd:466` — vendor-gap list named two labs absent from the post and omitted several present in it. → rewritten against the actual tables.
8. `index.qmd:264` — heading "Recognition on hard audio is where the cascade earns its keep" asserted a comparison the section never makes and partly refutes. → renamed.
9. `index.qmd:343` — the 150× headline compares the two number types the † note calls incomparable; "two thirds of listeners prefer" ≠ a 66% pairwise win rate. → both fixed.
10. `index.qmd:341` — "sits fifth" asserts a rank the overlapping CIs do not support, in a post that tells readers CIs are how you check exactly that. → hedged with the intervals.
11. `index.qmd:415` — same benchmark reported as both percentages and 0–1 fractions in adjacent sentences. → unified.
12. `index.qmd:485` — the lone RTFx figure had no source or retrieval date. → footnoted, with the comparison made explicit.
13. `index.qmd:455` — vendor table said "Plan-based" while the TTS table quoted \$49/1M chars. → reconciled, noting the latter is a normalisation rather than a published rate.
14. `make_figs.py:53` — `is_cascade` populated but never read; the baseline was distinguished only by tick text. → now emphasises the cascade's label.

PR #149, opened 2026-08-10
feat(posts): prompt evaluation pipelines and LLM-as-judge

project-delphi:
## Summary

New prose post: `posts/prompt-evaluation-pipelines/index.qmd`, aimed at engineers and technical PMs who ship LLM features but have never formalized how they test prompt changes.

- The spine: a prompt change can only be judged against a fixed test set and a scorer, and for the qualities that actually matter the scorer has to be another model — so the real problem is keeping that judge honest.
- Nine claim-style H2s covering the six pipeline stages, deterministic vs. model-graded scoring (with a contrast table), judge mechanics, the three judge patterns, an inline `## Caveat:` section on judge failure modes, calibration against human labels, and production realities (PR gating, growing the set from traces, offline pass ≠ launch signal).
- A support-assistant RAG example runs through the middle of the post and carries the one code block — a groundedness judge prompt emitting strict JSON.
- No `References` section: nothing in the post rests on a citable source, and judge-bias claims are stated qualitatively rather than with invented agreement percentages or benchmark numbers.

Prose-only post, so no `jupyter:` key, no venv, no `requirements.txt` — the judge prompt is a brace-free ```` ```text ```` fence, so `check_posts.py` classifies it as non-executing. Cover generated by the committed `src/make_cover.py` in the house style (`#4A3AA7`, "AI ENGINEERING" badge, triple-ring mark).

`docs/` is re-rendered and committed in the same PR per CLAUDE.md.

## Test plan

- [x] `make check-posts` passes — `OK (5 legacy posts grandfathered)`; new post added to neither `LEGACY_NO_ENV` nor `STALE_FREEZE_OK`.
- [x] `QUARTO_PYTHON="$(pwd)/.venv/bin/python" quarto render .` completes (84/84), `Output created: docs/index.html`.
- [x] `git status` after render shows only the new post dirs plus `docs/index.html`, `docs/listings.json`, `docs/search.json` — no `_freeze/` churn on other posts.
- [x] Rendered page verified: 6-row table, code block, 13 ToC entries, cover image present, no stray fenced-div markers in body text.
- [x] Post appears on the homepage listing (date-desc, top) with a 7 min reading time.
- [x] Served locally at `http://localhost:8000/posts/prompt-evaluation-pipelines/index.html` → HTTP 200.

Body length: 1216 words including headings (1132 excluding), slightly over the 1150 target ceiling; trimming further would have meant dropping required content.


project-delphi (commented):
## Automated review (`/code-review`, default level)

**Mechanics verified clean:**

- `scripts/check_posts.py` → `OK (5 legacy posts grandfathered)`. The ```` ```text ```` fence is correctly not matched by `CELL_RE` (which only matches ```` ```{lang} ````), so no kernel/`requirements.txt` is required.
- `docs/` is in sync with source — every distinctive sentence from the final `.qmd` is present in the rendered body.
- `docs/search.json`: 615 → 624 entries. 10 added (all this post), 1 removed — `voice-ai-architectures-2026#recognition-on-hard-audio-...`, an anchor that does not exist in that post's source. Stale hand-edited entry from `a097d5c`; this render correctly drops it. Remaining churn is entry reordering only.
- `docs/index.html`: the `src=` → `data-src=` flip on `posts/cursor_ai/cover.png` is Quarto's lazy-load threshold shifting by one as a row is inserted, not a broken thumbnail.
- `cover.png` byte-identical between `posts/` and `docs/`, 1200×630, matches what `src/make_cover.py` draws. `src/` correctly not copied into `docs/`, matching the other 10 posts with a `src/`.

### Critical Issues

None.

### Important Issues (Medium)

- **`index.qmd:111` — a groundedness judge cannot separate a retrieval failure from a generation failure.** The judge as specified sees only `{retrieved_chunks}` and `{candidate_answer}`. Retriever misses the chunk and the model answers from memory → `fail`. Retriever finds it and the model ignores it → also `fail`, indistinguishable. Separating the two needs a second signal (context relevance / retrieval recall) the post never introduces.
- **`index.qmd:146` — the merge gate contradicts the post's own noise caveat.** Line 117 says borderline cases flip between runs and "a one-point improvement can be noise"; line 146 then blocks merges on any groundedness drop, with no noise band or minimum effect size. A clean PR with three flipped borderline cases gets blocked by exactly the flakiness the caveat warns about.
- **`index.qmd:137` — "keeping only self-consistent verdicts" discards data non-randomly.** Dropping pairs where the judge contradicts itself removes precisely the close calls, so the surviving subset reports a win rate over a filtered, easier population. The inconsistency rate *is* the bias measurement and belongs in the reported number, not the discard pile.

### Suggestions (Low)

- **`index.qmd:130`** — "agreement with those labels is its accuracy" omits the class-imbalance trap: on a 90/10 pass/fail set, a judge that always says `pass` scores 90%. Also, at the recommended n=50 the 95% CI is roughly ±14 points, which can't support the one-point-noise standard set elsewhere.
- **`index.qmd:59`** — passing unit tests is a proxy (generated code can pass supplied tests and fail an untested edge case). An unusually strong proxy, but the current wording asserts identity with the target property.
- **`index.qmd:159`** — the closing pays off with "citation discipline", which is never established as a criterion; the narrow-judge list at 133 is "grounded, answered the question, right tone, correct format".

### Informational

- `src/make_cover.py:39` — `ruff` D406/D407 on the Google-style `Returns:` block, because `pyproject.toml` sets `convention = "numpy"`. Not introduced here — every existing `posts/*/src/make_cover.py` fails identically, and hooks aren't installed by default.

### Strengths

Repo invariants all hold: prose-only classification correct, `docs/` rendered in the same commit, no `_freeze/` churn on other posts, cover script a faithful copy of the house-style original, and the stale `search.json` anchor cleaned up as a side effect.

---

Addressing the three Medium findings and the three Low ones in a follow-up commit before merge.

PR #150, opened 2026-08-10
Post: wiring the groundedness judge to the Claude API

project-delphi:
## Summary

A short (~420 words of prose) code-first sequel to [#149](https://github.com/project-delphi/ml-blog/pull/149).

That post argued *why* LLM-as-judge sits at the centre of an eval pipeline, and asserted the harness should "parse `verdict` and treat a malformed answer as an error, not a score" — but its judge was a `text` fence ending in *"Reply with JSON only"*, which is a request the model is free to ignore. This post shows what makes that assertion true:

- **`messages.parse()` with a Pydantic model** — the format instruction leaves the rubric entirely and the parse step disappears. `unsupported_claims` is declared before `verdict` so evidence is generated before label.
- **A test set that's situated, per the repo's writing conventions** — six synthetic `(context, answer)` pairs with stated provenance, generating process, downstream cost, and why synthetic beats a scraped corpus here (agreement needs known labels).
- **The Batches API for the fan-out** — plus the two details that bite: the batch path takes the canonical `output_config.format` rather than `parse()`'s `output_format=`, and `extra="forbid"` is what makes the emitted schema legal; results arrive in any order, so cases key on `custom_id`.
- **The close** returns to the spine: agreement with hand labels is what licenses quoting a pass rate, and the one case the judge misses is the hallucination wearing the documentation's vocabulary.

**Cells are display-only** (`#| eval: false`, shared `blog-base` kernel). There are no API credentials on the authoring machine, and the post makes its point without a live call — so it adds no `.venv-<slug>`, no `requirements.txt`, and nothing re-executes on a project render. The two output blocks are labelled in prose as illustrative ("Illustratively:", "would plausibly give:"), never presented as captured runs.

## Test plan

- [x] `make check-posts` green before and after the render
- [x] `QUARTO_PYTHON=… quarto render .` — full project render; touched only `docs/index.html`, `docs/listings.json`, `docs/search.json` and the new post, confirming `freeze: auto` re-executed nothing
- [x] All four `{python}` cells parse under `ast.parse`, individually and concatenated; every one carries `#| eval: false`
- [x] `grep -c cell-output docs/posts/claude-api-eval-pipeline/index.html` → 0 (no cell displays executed output)
- [x] Verified against `anthropic` 0.121.0 that `messages.parse`, `messages.batches.results`, and both `MessageCreateParamsNonStreaming` / `Request` import paths exist
- [x] Verified in Pydantic that `extra="forbid"` is exactly what emits `additionalProperties: false` (absent without it) — the claim the post makes about schema legality
- [x] Eyeballed the rendered page in-browser: cover, ToC reconstructs the argument, code unfolded, cross-post link resolves


project-delphi (commented):
## Automated review — round 1 (`review` skill, multi-agent)

### Verified clean
- `make check-posts` passes; `_freeze` hash matches the source md5 exactly, no stale-freeze drift.
- `docs/` is purely additive: `search.json` 624 → 629 entries, zero removals; `listings.json` 83 → 84; no other post's HTML touched. Cover PNG byte-identical between `posts/` and `docs/`.
- Cross-post link `../prompt-evaluation-pipelines/index.qmd` resolves.
- API surface correct against current docs: `claude-opus-5`, `messages.parse(output_format=…)` → `.parsed_output`, `output_config.format` for the batch path, `Request` / `MessageCreateParamsNonStreaming` import paths, unordered batch results.
- Ran the `Verdict` model under Pydantic 2.13 and confirmed the post's load-bearing claim: `ConfigDict(extra="forbid")` does emit `additionalProperties: false`. Arithmetic consistent: 4/6 = 0.67, 5/6 = 0.83.

### Findings (all three real, all fixed in b318c4d)
1. **`max_tokens=1024` budgets thinking and output together** (two sites). Thinking is on by default on Claude Opus 5 — a judge decomposing a multi-claim answer could truncate and return `stop_reason: "max_tokens"` with malformed JSON, i.e. the harness reporting a budget shortfall as a malformed model reply. That is exactly the failure the post claims typed verdicts eliminate.
2. **`next(...)` without a default** turns a text-block-less success into a bare `StopIteration` that escapes the loop, bypassing the `custom_id`-tagged error and naming neither case nor reason.
3. **`judge()` annotated `-> Verdict` but `parsed_output` is `Optional`** — `None` on refusal or truncation would flow downstream silently. (Flagged lower-confidence by the reviewer; verified directly against the installed SDK: `parsed_output: Optional[ResponseFormatT] = None`.)

---

## Automated review — round 2 (after the round-1 fixes)

### Verified clean
`make check-posts` passes, frozen hash matches, rendered HTML reflects the final source, cover committed in both trees, cross-post continuity claim accurate (that post's rubric does end in `Reply with JSON only:` and uses the same field names). API surface re-checked and correct, including the 50% batch discount.

### Findings (all three fixed in 5b8201e)
1. **medium — the worked example contradicted the conclusion.** The illustrative `judge()` output showed the judge *catching* `Refunds are processed within 3 business days.` — verbatim the `refund-window-invented-sla` answer — while the results block and the close both build the payoff on the judge *missing* that case. Fixed by making the illustrative output a different fabrication (one of the two the judge does catch), so the narrative holds: obvious inventions are caught, the one wearing the documentation's vocabulary is not. Arithmetic unaffected.
2. **medium — the batch loop didn't cover the truncation case the prose claimed it did.** A case cut off mid-JSON returns a *partial* text block, not a missing one, so `text is None` was False, the named `RuntimeError` never fired, and `model_validate_json` raised a bare `ValidationError` with no `custom_id` — the "dies at case 37 of 200" failure the post opens with, reintroduced on the path it says eliminates it. Guard now checks `stop_reason != "end_turn"`.
3. **low — `max_tokens=8192` thinner than the post's own argument warrants.** Raised to the documented non-streaming default of 16000.

---

**Disposition:** no Critical findings in either round. All six findings were real and all six are fixed; rounds 1 and 2 each re-rendered and re-verified (`check-posts` green, zero `cell-output` blocks, no other post's output touched). Merging.

PR #155, opened 2026-08-12
New post: explainability is not causality, and a CLAUDE.md preview-link rule

project-delphi:
## Summary

A short bridge post between the two existing deep-dives (`explainability-localization` and `causality-from-toddlers-to-do-calculus`), arguing that a feature-attribution result describes an input-output association inside a fixed model, while a causal claim is bought by intervening on the world. Mistaking the first for the second is the common failure this post is aimed at.

- Prose-only post, so no kernel pin, no `requirements.txt` and no `_freeze/` record — `scripts/check_posts.py` exempts code-free posts.
- Three hand-written SVG diagrams (localize vs. intervene, Pearl's ladder, wolf/snow contrast), each with an opaque white background rect so Chrome's auto-dark mode can't invert them into illegibility.
- House-style `cover.png` from `src/make_cover.py`.
- Figure 11 of Ribeiro et al. (2016) reproduced under fair use with attribution and a link to the paper. The arXiv posting grants no redistribution licence and the KDD version is ACM-copyrighted; including it was an explicit call by the repo owner, not a side effect of the render. Flagging it here so the decision is on the record.
- `CLAUDE.md`: the preview link now goes last in the reply, and that rule is reconciled with the existing server-kill rule (after a merge there's no localhost server, so that reply ends with the published URL).

## Test plan

- [x] `make check-posts` — `OK (5 legacy posts grandfathered)`, count unchanged
- [x] `QUARTO_PYTHON=... quarto render .` — exit 0, no errors; `freeze` held, so only the listing pages changed
- [x] Rendered page checked in a browser: two study panels side by side, all three SVGs legible, 12-row table renders as a real `<table>`, TOC reconstructs the argument
- [x] Both cross-links resolve to existing rendered pages
- [x] Prose 600 words, inside the 400-600 target
- [x] A `/code-review medium` pass was run and its findings applied — see below

## Review findings already fixed

- The husky/wolf training set was **constructed**, not a correlation the authors stumbled on: 20 images hand-picked so every wolf had snow and no husky did. The post said the latter; verified against the PDF and corrected.
- The panel (b) caption claimed the animal was fully masked out. The image beside it visibly retains the collar/chain and part of the chest — caption corrected.
- Clarified why ablation/patching earns a causal claim about the network while an ML counterfactual doesn't: intervening on internals vs. searching inputs.
- Widened the cover's category pill, which was cramping a 26-character badge.


project-delphi (commented):
## Review — `/code-review medium`

**Scope:** `git diff main...HEAD` — 2 commits: the new post `posts/explainability-vs-causality/` (qmd, cover script, 3 SVGs, 2 reproduced PNGs), its rendered `docs/` output, and a 4-line CLAUDE.md addition. Working tree clean.

### Verification done
- `make check-posts` passes (no code cells, so no kernel/`requirements.txt`/`_freeze` obligation — correctly a tier-3 post).
- `docs/` in sync with source: the post's HTML, cover, SVGs and both PNGs all copied; `search.json` gained exactly the 8 new entries and lost none (647 → 655, no other post's index perturbed); `docs/index.html` listing re-indexed correctly.
- Cross-links `../explainability-localization/` and `../causality-from-toddlers-to-do-calculus/` both resolve to real post directories, matching the existing bare-directory link convention.
- Factual check of the LIME claims against Ribeiro et al. (2016) §6.4 / Fig. 11: 20 hand-picked training images, 10/27 → 3/27 trusting, 12 → 25 naming snow — all correct. The two PNGs are genuinely Fig. 11(a) and 11(b).
- Rendered figures, image refs, subtitle and cover all render; `make_cover.py` is the standard per-post copy of the house-style script (duplication is the established convention here, not a defect).

### Critical Issues
None.

### Suggestions (3, all low severity)
- `posts/explainability-vs-causality/index.qmd:37` — Fig. 11 is reproduced from the arXiv/KDD paper, which carries no CC license (arXiv's non-exclusive distribution license plus ACM copyright). The credit line says "for commentary", which is the right instinct, but this is redistribution of a copyrighted figure on a public site; worth a conscious call rather than an implicit one.
- `posts/explainability-vs-causality/index.qmd:34` — caption says the explanation kept "snow, and the collar. Almost all of the animal is gone." The retained superpixels also include the dog's dark snout/mouth region, which is part of the animal. Minor overstatement of an otherwise accurate point.
- `CLAUDE.md:31` — instructs ending the post-merge reply with the Pages URL. GitHub Pages redeploys asynchronously after the merge commit lands, so that URL 404s for a minute or two at the moment it is handed over. Worth a half-sentence that it goes live once Pages rebuilds.

### Strengths
No correctness bugs found. Provenance claims about the LIME experiment check out against the primary source, and `docs/` is consistent with source with no collateral churn.

### Disposition
The copyright note is a conscious, documented decision (see PR description). The other two are being fixed in a follow-up commit on this branch before merge.

PR #160, opened 2026-08-13
feat(posts): add eigendecomposition primer

project-delphi:
## Summary

- Adds `posts/eigendecomposition/` — a friendly entry point for eigenvectors and eigenvalues. The blog currently covers them only inside `matrix-factorizations`, which derives $Av = \lambda v$ from the Rayleigh quotient on a symmetric matrix; there was nothing for a reader meeting the idea for the first time.
- Differentiates deliberately: one **non-symmetric** 2×2 matrix, one figure, plain language, 399 words of prose. Links onward to matrix-factorizations for the variational derivation rather than repeating it.
- Tier-1 post on a new `.venv-eigen` / `eigen-blog` kernel, so `kernels-stub` in the `Makefile` gains that name — without it a fresh clone cannot render *any* page, since kernelspecs resolve before `_freeze/` is consulted.
- Cover image in house style via `src/make_cover.py`, plus `requirements.txt` freeze and committed `_freeze/` + `docs/`.

Two things worth a reviewer's eye:

- numpy 2.5 returns complex dtype from `eig` for any non-symmetric matrix, so the demo takes `.real` (with a comment) — otherwise output reads `5.00+0.00j`.
- Because `A` is not symmetric, the eigenvector arrows land *on* the image ellipse but *not* on its axes — those belong to the singular vectors. The post says so inline; the claim was checked against the rendered figure, not assumed.

## Test plan

- [x] `make check-posts` → `OK (5 legacy posts grandfathered)` (kernel pinned, `requirements.txt` present, freeze hash matches source)
- [x] `quarto render posts/eigendecomposition/index.qmd` with `QUARTO_PYTHON=.venv-eigen/bin/python` — executes clean
- [x] Full-site `quarto render .` — only `docs/index.html`, `listings.json`, `search.json` changed outside the new post, so no other post re-executed
- [x] Rendered figure inspected: equal aspect, arrows 5:2, arrows visibly off the ellipse axes as the text claims
- [x] Reconstruction cell prints `max |A - P D P^-1| = 8.88e-16`
- [x] Prose word count 399 (excludes YAML, code, table, captions)


project-delphi (commented):

Reviewed as a single careful diff pass. Executable claims verified by running the post's own `.venv-eigen` interpreter; freeze hash checked against source; both committed images inspected.

### What checks out
- `_freeze` hash `ad4ea47c…` equals `md5(posts/eigendecomposition/index.qmd)` — freeze is in sync, so a project render won't re-execute it.
- `eigen-blog` added to `kernels-stub`; `.venv*` already gitignored; `requirements.txt` restore instructions are correct.
- Math is right: trace 7, det 10, χ(λ)=λ²−7λ+10; eigenvectors (1,1)/(1,−2); reconstruction residual ~1e-16.
- `docs/` is consistent: `search.json` gains 4 entries / 1 page with no losses, `listings.json` gains the post, no other post re-executed.
- Figure and cover render correctly (equal aspect, arrow tips on the ellipse, no clipping at ±6).

### Findings (4 — no Critical, no correctness bugs in shipped output)

**1. `posts/eigendecomposition/index.qmd:47` — prose/output mismatch at the verification moment.** The text says "`A @ v` is `v` with every number multiplied by the same amount — (1,1) by 5, (1,−2) by 2", but the printed cell shows `eigenvector = [0.707 0.707] A @ v = [3.536 3.536]`. numpy returns unit-norm eigenvectors, so a beginner sees neither a 1 nor a 5 and cannot check the claim against what's on screen.

**2. `posts/eigendecomposition/index.qmd:32` — the comment attributes the complex dtype to asymmetry.** Verified on numpy 2.5.2: `eig` returns `complex128` for `[[4,1],[1,3]]` (symmetric) and `[[2,0],[0,3]]` (diagonal) too. The dtype is unconditional in numpy 2.x.

**3. `posts/eigendecomposition/index.qmd:145` — the closing caveat names only defective matrices.** The far more common failure is genuinely complex eigenvalues: a rotation like `[[0,-1],[1,0]]` has no real eigendirection at all (~29% of random real 2×2s), versus a measure-zero set of defective ones. Combined with the unconditional `vals.real`, a reader who swaps in a rotation silently gets real parts printed as if they were the answer.

**4. `posts/eigendecomposition/index.qmd:33` — `vecs * np.sign(vecs[0])` zeroes any eigenvector with a zero first component**, since `np.sign(0) == 0`. Harmless for this fixed `A`, but this is a copy-pasteable snippet: for `[[1,0],[0,2]]` a column becomes all zeros, making `P` singular.

### Minor (not filed)
Legend renders two visually identical purple entries distinguished only by the λ text; and "come close to" the ellipse axes is 5.2° off for λ₁ but 13.3° for λ₂, so "close" does more work for the second arrow.

Nothing here blocks the merge; 1 and 3 are the ones worth fixing before publishing, since this is an explicitly beginner-facing post.

project-delphi:
Applied fixes for all four review findings in 49b347f:

- **1 (prose/output mismatch)** — eigenvectors are now rescaled to lead with a 1, so the cell prints `[1. 1.] → [5. 5.]` and `[1. -2.] → [2. -4.]`. The reader can now check the sentence against the screen, which was the whole point. The trailing "(1,1) by 5, (1,−2) by 2" clause in the prose became redundant and was dropped.
- **2 (wrong comment)** — confirmed independently on this venv's numpy 2.5.2: `eig` returns `complex128` for symmetric and diagonal input too. Comment now says numpy 2.x returns complex dtype unconditionally.
- **3 (missing failure case)** — the closer now names rotations, which turn every direction and have no real eigenvector at all, alongside defective matrices.
- **4 (`np.sign(0) == 0`)** — the sign normalisation is gone entirely, replaced by the rescale above; the comment flags that a zero first entry would need another pivot.

Re-verified: `make check-posts` passes, freeze hash back in sync, figure re-inspected (unchanged — the plot normalises to unit length itself), reconstruction still `8.88e-16`, prose 399 words.

PR #163, opened 2026-08-15
docs: add a prose style guide for posts

project-delphi:
## Summary

- Posts have been reading wordy and hard to follow: they reach for code, diagrams and jargon before explaining in plain language what problem is being solved, and lean on long multi-clause sentences.
- Adds `notes/blog-style.md` — sentence- and paragraph-level rules (warm-up prose before mechanics, short sentences, active voice, everyday words over Latinate ones) plus two calibration examples (cross-validation, the bootstrap) that fix the target tone.
- `CLAUDE.md` gains a single pointer line under the existing `## Writing conventions`, so the always-loaded instructions stay short and the style rules can grow in their own file.

The guide lives in `notes/` rather than `docs/`: `docs/` is Quarto's `output-dir` and is deleted and rebuilt on a project render, while `_quarto.yml` already excludes `notes/` from rendering. The existing `## Writing conventions` rules are untouched — they cover post-level argument shape, the new file covers prose.

## Test plan

- [x] `make check-posts` passes — no post source touched, no freeze drift.
- [x] `git status` shows only `CLAUDE.md` and `notes/blog-style.md`; no `docs/` churn, so no re-render needed.
- [x] `grep -n 'notes/' _quarto.yml` confirms the `!notes/` render exclusion, so the guide is never published to the site.
- [ ] Not verified: whether the guide actually makes future posts less wordy — that shows up on the next post written against it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #164, opened 2026-08-15
docs: relocate and scope the blog style guide

project-delphi:
## Summary

Follow-up to #163, addressing the four findings its review raised.

- **Moved `notes/blog-style.md` → `STYLE.md`** at the repo root. `notes/` is documented in `CLAUDE.md` as "a scratch area, not published content", so a normative guide living there was one cleanup away from deleting itself and leaving the CLAUDE.md pointer dangling. The root is not scratch, and `STYLE.md` is still not rendered — `_quarto.yml` globs only `*.qmd`/`*.ipynb`. The now-empty `notes/` directory is gone, so CLAUDE.md's description of it is accurate again.
- **Scoped the guide off the legacy posts.** A new `## Where these rules apply` section states that the rules govern new posts and posts that already carry a venv + kernel + `requirements.txt`, and explicitly do not license rewriting the five `LEGACY_NO_ENV` posts — any prose edit there invalidates the md5-keyed `_freeze/` record and makes the next project render try to execute a post with no environment.
- **Deduplicated the code/figure rule.** `CLAUDE.md` and the guide both said how much prose a code block is owed, and disagreed (2–4 sentences of warm-up vs. one sentence before and after). Rather than rank them, `CLAUDE.md` gives up its code/figure sentence and keeps the section-seam rule; `STYLE.md` is now the sole authority and states the graded rule directly — warm-up on the first reach for a new idea, one sentence for routine blocks after that.
- **Stopped the sentence rule from flattening the house voice.** "Split anything with more than one subordinate clause" would have rewritten a corpus built on em-dash-joined sentences. It now targets the sentence you have to read twice, keeps em-dash glosses, and yields to an existing post's register.

Net effect on `CLAUDE.md`: one pointer line in (#163), one sentence out — 17,540 bytes vs 17,654 on `main`.

## Test plan

- [x] `make check-posts` passes — no post source touched, no `_freeze/` drift.
- [x] `git diff main --stat` shows only `CLAUDE.md` and the `notes/blog-style.md → STYLE.md` rename; no `docs/` churn, so no re-render needed.
- [x] `grep -n 'blog-style'` finds no surviving reference to the old path outside built `docs/`.
- [x] `_quarto.yml` render globs confirmed to exclude root markdown, so `STYLE.md` is not published to the site.
- [ ] Not verified: whether the guide actually makes future posts less wordy — that shows up on the next post written against it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ done
 
 ## Writing conventions
 
-For blog posts, read and follow `STYLE.md` before writing.
+For blog posts, read and follow `STYLE.md` before writing or editing one.
 
 **Give every post a spine.** A post is one argument, not a pile of sections. Name the core idea in a sentence before writing; every section advances it.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ done
 
 ## Writing conventions
 
-For blog posts, read and follow `notes/blog-style.md` before writing.
+For blog posts, read and follow `STYLE.md` before writing.
 
 **Give every post a spine.** A post is one argument, not a pile of sections. Name the core idea in a sentence before writing; every section advances it.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -51,7 +51,7 @@ For blog posts, read and follow `STYLE.md` before writing.
 - **Open on the core idea and the stakes** — the question, why it's non-obvious, what changes once the reader knows. No "in this post we will…" preamble.
 - **Headings state claims, not topics** — `## Efron's bootstrap is a weighted bootstrap in disguise`, not `## Background`, so the ToC reconstructs the argument.
 - **Each section earns the next.** If two could swap without damage, merge or cut one. `###` is for steps within one idea, not new ideas.
-- **Stitch every seam.** A section's first sentence links back to the previous result or the core idea; its last names the unresolved thing the next section answers — the gap, not the mechanics ("next, some code"). Same for code and figures: a sentence before saying what it will show, one after saying what happened. If an opening sentence reads identically with the previous section deleted, the seam isn't stitched.
+- **Stitch every seam.** A section's first sentence links back to the previous result or the core idea; its last names the unresolved thing the next section answers — the gap, not the mechanics ("next, some code"). If an opening sentence reads identically with the previous section deleted, the seam isn't stitched.
 - **Close by returning to the core idea** — restate the opening claim now that it's earned, what it buys, where it stops holding. Not a summary of sections.
 - **Caveats inline**, as a short section where the objection occurs (`## Caveat: the uniform is the posterior, not the prior`) — not a "Limitations" bin at the end.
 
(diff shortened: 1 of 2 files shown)


project-delphi (commented):

Scope: `git diff main...HEAD` — 2 files, docs-only (`CLAUDE.md` edit + `notes/blog-style.md` → `STYLE.md` rename with content changes). No `.qmd`/`.ipynb` source changed, so the "re-render `docs/` in the same PR" rule doesn't bite here.

### Findings (4) — no Critical issues

1. **`STYLE.md:5` — Medium. Applicability test excludes safe-to-edit posts.** The scope sentence ("edits to posts that already carry their own venv, pinned kernel, and `requirements.txt`") uses the tier-1 environment convention as the test, but that excludes the four tier-2 Claude-API posts (`claude-api-eval-pipeline`, `langgraph-vs-llamaindex`, `messages-api-streaming`, `structured-json-with-claude`), the eight `.ipynb` posts, and the prose-only `STALE_FREEZE_OK` posts (`pinecone-vs-weaviate`, `pyspark`) — none of which have a `requirements.txt`, and none of which are fragile. An agent asked to tighten prose in `messages-api-streaming` would find the post outside the governed set and either decline or fall back to unguided editing. The actual hazard is `_freeze/` md5 invalidation, which applies only to `LEGACY_NO_ENV`.

2. **`CLAUDE.md:166` — Low.** Still describes `notes/` as a scratch area excluded from rendering. The rename removed the only file in it and the directory no longer exists on disk, so the exclusion is now vestigial rather than load-bearing, with no signal saying so.

3. **`_quarto.yml:7` — Low.** The `- "!notes/"` render exclusion now matches a directory that doesn't exist. Quarto tolerates the no-op negation and the site still renders (the render list is `*.qmd`/`*.ipynb` only, which is also why the new root `STYLE.md` is correctly not picked up into `docs/`). Dead config that will outlive its reason.

4. **`CLAUDE.md:54` — Low.** The hunk deletes the code/figure sentence, and the replacement graduated rule lives only in `STYLE.md`. CLAUDE.md is loaded unconditionally; `STYLE.md` is read only when an agent follows the line-47 pointer. A small prose edit that doesn't trip "read STYLE.md before writing" now proceeds with no code/figure framing rule in context at all.

**No correctness issues in the retained `STYLE.md` content** — the self-referential claim "both calibration examples below do it" was checked against the two examples, and both do use em-dash gloss clauses.

### Fixed before merge (`e6ddbf8`)

- **#1** — the carve-out is now the five `LEGACY_NO_ENV` posts and nothing else; the scope line reads "new posts and edits to existing ones, with one exception."
- **#4** — the CLAUDE.md pointer now reads "before writing or editing one", so a prose edit can't skip the guide. Closes the gap without reintroducing a duplicate rule in CLAUDE.md.

### Not fixed (deliberate)

- **#2 and #3** — `notes/` is recreatable at any time and the `!notes/` exclusion is what makes it a valid scratch area, so the config and its description still describe something true. Removing them would be churn for a directory someone may well want back.

PR #165, opened 2026-08-15
style(posts): rewrite the Kendall shape post against STYLE.md

project-delphi:
## Summary

A prose pass over `posts/kendall-shape-analysis/index.qmd` — the most recent post — against the style guide added in #163/#164. Code cells are untouched; every change is prose.

- **The opening asserted its premise cold.** It led with the workshop anecdote and the Kendall genealogy, so the reader met the method before any reason to care. Now opens on photographing the same leaf twice, names shape and the stakes (the space is curved, your next tool assumes flat), then keeps the genealogy beat.
- **First principles before the term.** Landmarks are derived — points you can find again on every specimen — before being named; centroid size and the Frobenius norm likewise.
- **The Fréchet mean was used in code and never explained.** Added the derivation: an arithmetic average falls off the sphere, so take the point minimising total squared distance along the space.
- **GPA arrived as a name plus a mechanism.** The chicken-and-egg problem now comes first, and the iteration answers it.
- **Anatomy was assumed.** Disc, cup and rim were jargon on first use; there's now a plain warm-up on the optic nerve head and why glaucoma is read off it.
- **Two sections opened on a non-prose block** — a bullet list of formulas, and a callout. Both now have prose lead-ins. The log and exp maps are explained as operations before being named.
- Long stacked-clause sentences split throughout, and the close returns to the leaf from the opening.

## Test plan

- [x] `quarto render posts/kendall-shape-analysis/index.qmd` with `QUARTO_PYTHON` set to `.venv-kendall` — all 9 cells execute clean (the prose edits invalidate the `_freeze` md5, so it re-ran).
- [x] `make check-posts` passes.
- [x] Every number in the prose re-verified against the freshly executed output, not the old text: residuals 0.060–0.077 ("0.06 to 0.08"), within-monkey distances 0.028–0.289 ("0.03 to 0.29"), tangent rank 8, 87.8% ("88%"), PC1 gap 0.12 sd, PC2 0.61 sd at p = 0.21.
- [x] `docs/` re-rendered and committed alongside the source.
- [x] Rendered page served locally and read through.


project-delphi (commented):
## Multi-agent review (`/code-review`, effort: medium)

Scope: `posts/kendall-shape-analysis/index.qmd` plus the re-rendered `docs/` and `_freeze/` (branch `rk/kendall-style-guide-pass` vs `main`). The change is a prose/style pass — no code cells were touched. Every numeric claim was verified against the rendered output, along with render hygiene.

### Verified clean
- All quoted numbers match the committed render: residuals 0.060–0.077 ("0.06 to 0.08"), 87.8% ("88%"), PC1 gap 0.12 sd, PC2 gap 0.61 sd, p = 0.21, tangent rank 8, within-monkey distances 0.028–0.289 ("0.03 to 0.29").
- `docs/` and `_freeze/` were re-rendered in the same branch; freeze hash matches the new source, and `make check-posts` passes.
- Dimension arithmetic (`km - m - 1 - m(m-1)/2` = 8 for k=5, m=3) and the Gower SVD rotation statement are correct as written.

### Critical Issues
None.

### Important Issues

**Medium — `posts/kendall-shape-analysis/index.qmd:336`: new paragraph contradicts the post's own earlier code.** It states "the ordinary arithmetic average is not available on a curved space — average points spread over a sphere and the answer falls somewhere inside it, off the surface entirely." But the pencil GPA a few sections earlier does exactly that: `new_mean = centre_and_scale(aligned.mean(axis=0))` takes the arithmetic mean and re-projects it onto the sphere. A reader who followed that code will read this as either "the earlier section was wrong" or "these are two different means with no explanation." The distinction is real (Procrustes mean vs. Fréchet mean under the quotient metric) but the new text asserts unavailability rather than making it; it needs a clause acknowledging that projecting the arithmetic mean back is the cheap approximation and that `geomstats` does the intrinsic version.

### Suggestions

**Low — `posts/kendall-shape-analysis/index.qmd:338`: the Fréchet mean is described as a medoid.** "take the point whose total squared distance to all the others is smallest" reads as selecting one of the 22 specimens, which is the medoid, not the Fréchet mean. The very next figure plots the mean in black as a point distinct from all 22, so the description contradicts the figure. Fix: "the point — not necessarily one of the data points — minimising total squared distance to them all."

**Low — `posts/kendall-shape-analysis/index.qmd:470`: off-by-one cross-reference.** "That is the dimension count from two sections ago" — the count is derived in `### Fifteen numbers, eight dimensions of shape`, which is the immediately preceding section. Should be "the previous section" / "one section ago".

---

All three findings are addressed in the follow-up commit on this branch.

PR #167, opened 2026-08-15
style(posts): rewrite the eval-pipeline post against STYLE.md

project-delphi:
## Summary

A STYLE.md pass over `posts/claude-api-eval-pipeline/index.qmd`, in the same shape as the Ripley (`fa6a95a`) and Kendall (`1249232`) passes.

- **The opening now stands alone.** It used to start "You have the groundedness judge from [the previous post]", so a reader arriving from search met *groundedness*, *rubric* and *LLM-as-judge* undefined. The new opening derives the framing from the reader's own problem — output quality you can't write an `assert` for — and names each term after explaining the idea. The previous post is now further reading, not a prerequisite.
- **Warm-ups before the mechanics.** Structured outputs, Pydantic, and the Batches API each got one sentence before a code block; STYLE.md asks for 2–4 the first time a section reaches for a new idea.
- **Jargon glossed on first use**: JSON schema, Pydantic model, `custom_id`, `stop_reason`, "the strict subset" / `additionalProperties: false`, pass rate, agreement, calibration set.
- **Split the sentences you have to read twice** — the `max_tokens` paragraph (three clauses across an em-dash and a colon) and the `succeeded`-is-a-transport-verdict paragraph.
- **The data is situated properly** per CLAUDE.md: who wrote the six cases, what they stand in for, what "context" and "label" mean, what a wrong verdict costs a customer, and why synthetic beats a real dataset (judge–human agreement needs known labels).
- **Caveats moved inline.** "Six cases is a demonstration" and "none of it measures the retriever" left the closing for a `## Caveat:` section placed where the objection occurs, so the close can return to the core claim.
- **Fixed a number.** The new prose initially read the pass rate as "two of three"; the worked output says `0.67` over six cases, i.e. four of six. Corrected before commit.

Prose only — all six fenced blocks and the frontmatter are byte-identical to `main`. The code was checked against the current Claude API surface (`messages.parse(output_format=...)`, `output_config.format` on the batch path, `extra="forbid"` for the strict subset, 50% batch pricing, `claude-opus-5`) and needed no change.

## Test plan

- [x] `make check-posts` → `check-posts: OK (5 legacy posts grandfathered)` (the freeze record was regenerated by the project render, so it is not stale)
- [x] Full project render with `QUARTO_PYTHON="$(pwd)/.venv/bin/python" quarto render .`
- [x] `git diff --stat` limited to the five expected files — no other post re-executed
- [x] Scripted check that all six fenced code blocks and the YAML frontmatter are unchanged from `HEAD`
- [x] `codespell` clean on the edited file
- [x] Rendered page read in `quarto preview`: opening makes sense with the previous post unread, every `##` heading states a claim, ToC reconstructs the argument

Not run: `pre-commit run --all-files` rewrites ~235 unrelated files repo-wide (pre-existing whitespace/EOF/ruff debt). Those were kept out of this PR.


project-delphi (commented):
## Automated review — `/code-review 167` (effort: medium)

**Findings: none.** Reviewed hunk by hunk; no correctness issues worth submitting.

Since this is a prose-only diff, "bugs" here are factual and internal-consistency errors. What was checked:

**Numbers reconcile end to end.** 3 grounded / 3 fabricated cases → labels 3 pass, 3 fail. The judge passes the invented-SLA case (the one disagreement), so 4 of 6 passes = `0.67` ✓ and agreement 5 of 6 = `0.83` ✓. The new closing prose ("four of six", "five of six") matches the worked output block exactly — the "two of three → four of six" correction noted in the PR body did land. The hypothetical "agrees eight times in ten → one in five wrong → pass rate off by up to twenty points" is arithmetically sound.

**Prose still matches the untouched code.** "Two of the six, one of each kind" agrees with the `CASES` block (two shown, plus `# ...four more, three grounded and three ungrounded in total`); "an invented processing time" matches the `refund-window-invented-sla` case's "processed within 3 business days", which the customer-cost sentence then reuses.

**API claims verified against the current Claude API surface** (loaded the `claude-api` skill rather than answering from memory): `messages.parse(output_format=Pydantic)` vs the batch path's `output_config.format` + raw schema ✓; `extra="forbid"` → `additionalProperties: false` required by the strict subset ✓; Batches at 50% of list price, results in arbitrary order keyed by `custom_id` ✓; `end_turn` as the only clean stop reason, with refusals surfacing via `stop_reason` under a `succeeded` transport result ✓. The rewritten `max_tokens` paragraph — "caps the reasoning the model does before answering *and* the reply itself, drawn from one budget" — is correct for `claude-opus-5`, where thinking is on by default and `max_tokens` bounds thinking plus response text together.

**Build artifacts are in sync.** The `_freeze/.../html.json` `hash` equals the md5 of the new `index.qmd`, so the freeze record is not stale; `docs/posts/claude-api-eval-pipeline/index.html` contains the new prose including the new `## Caveat:` ToC entry; churn in `docs/index.html` and `docs/search.json` is mtime/reading-time only, with no other post re-rendered.

### Not a finding, flagged for transparency

The claim that exhausting the budget mid-JSON makes `parse()` raise (`ValidationError`) while exhausting it inside reasoning yields `parsed_output is None` could not be verified — the SDK is not installed in any venv here, and the skill docs don't specify which path a truncated parse takes. This assertion is carried over from `main` (only reworded), so it is not introduced by this PR. If it's ever worth pinning down, that's the line to test.

**Verdict: 0 Critical, 0 Important, 0 Suggestions.** Clear to merge.

PR #169, opened 2026-08-18
feat(posts): add the kite shape analysis post

project-delphi:
## Summary

A new post at `posts/kite-shape-analysis/`, built from five photographs of one kite taken at Villa de Leyva, Boyacá on 15 Aug 2026.

- **Why the post exists.** The five frames are of one rigid object nine seconds apart, so the correct shape difference is zero. Kendall shape analysis reports Riemannian distances up to 0.33 between them. The failure is not a bug: it quotients out the *similarity* group, and what a camera adds is perspective, which is not in that group.
- **Two independent pipelines** over the same hand-digitised `landmarks.csv` — a from-scratch Python GPA in the post, and R's `shapes` package (`procGPA`, `riemdist`, `tpsgrid`) in `src/shapes_analysis.R`. They agree to five decimal places, which is the point of running both.
- **The conclusion is narrower than the one I set out to write.** Quotienting out the affine group accounts for 38% of the variation, not the 74% a first (broken) attempt reported. A jitter simulation puts the remainder close enough to hand-digitising noise that the sail's billow cannot honestly be claimed from six landmarks. The post says so.
- **Two traps documented rather than hidden**: an iterative affine Procrustes collapses configurations onto a line while keeping unit size (caught by an aspect-ratio check, called out in the post), and a plain DLT homography minimises algebraic error and can score *worse* than the affine fit nested inside it.
- R runs ahead of the render with its figures committed, so this post builds on a machine with no R at all.

## Test plan

- [x] `quarto render posts/kite-shape-analysis/index.qmd` — clean, all 10 cells execute
- [x] `quarto render .` — full site render clean (the one fenced-div warning is pre-existing, from `posts/working-with-quarto`, untouched here)
- [x] `python3 scripts/check_posts.py` — OK; `kite-blog` added to `kernels-stub` in the `Makefile`
- [x] Every number in the prose checked against the rendered output
- [x] Python GPA vs R `procGPA` cross-checked: Procrustes SS 0.11100 / affine 0.06839 from both
- [x] All five landmark overlays inspected by eye — every point on the rigid sail, none on a streamer
- [x] `pre-commit` run scoped to the new files: `black`, `ruff`, `pyupgrade` pass; `add-trailing-comma` applied and both rewritten scripts re-run clean. `mypy` import-stub errors are pre-existing repo-wide (identical on `posts/kendall-shape-analysis/src/make_cover.py`) — it runs against the dependency-free base venv
- [x] `docs/` re-rendered and committed alongside the source


project-delphi (commented):

The pipeline was re-run in `.venv-kite` against the committed `landmarks.csv`. Every number in the prose reproduces exactly and matches `r_results.json` (similarity SS 0.11100, affine SS 0.06839, 38.4%, spread 31.6°, ladder 12.52/8.23/6.69, largest riemdist 0.326). `make check-posts` passes and the freeze is in sync. The problems are in two load-bearing methodological choices and one latent code bug.

### Critical / major

**1. `index.qmd:485` — the noise floor is computed from one arbitrarily chosen base frame, and the headline conclusion flips if you pick a different one.** At σ=5 the mean affine SS is 0.0287 with `X[4]`, but 0.0614 with `X[0]` and 0.0617 with `X[1]` (frames 1 and 2 are the flattest sails, so affine whitening amplifies jitter there). Observed affine SS is 0.0684. So "jitter of ±8 pixels alone produces 0.074 … so there is probably *something* left" becomes "σ=5 jitter on frame 1 already produces 0.061, i.e. essentially all of it" under an equally defensible base frame. The similarity claim survives but its margin also moves (3.8× the frame-5 floor, 2.8× the frame-1 floor — not uniformly "roughly four times"). Average over all five frames, or at minimum report the spread.

**2. `src/digitize_landmarks.py:88` — frame 5's `armA`/`armB` labels are deliberately swapped, disclosed only in a comment in a script the render never runs.** It sets the headline number: undo the swap and similarity SS goes 0.111 → 0.855, frame 5's residual 0.117 → 0.791, and the viewpoint share 38.4% → 91.7%. The post spends a whole section (`index.qmd:154`) insisting "point 3 must be the same physical spot on the kite in every photograph" and never says that in one of five frames it deliberately isn't. This belongs in the post.

### Important

**3. `src/digitize_landmarks.py:110` — stale `signed_area` docstring.** It claims the sign flips between frames and that this is why downstream fits allow reflection. Neither is true of the committed data: after the swap all five signed areas are negative, and `rotate_onto` (`index.qmd:254`) explicitly *forbids* reflection. It contradicts the sibling comment 20 lines above.

**4. `index.qmd:336` — real bug in `fit_similarity`, latent on this data.** The reflection branch fixes the rotation matrix but leaves the scale as `sv.sum()`; under reflection it must be `sv[0] - sv[1]`. On a mirrored target it returns rms 176.65 where the optimum is 160.86. It never fires here (all frames share an orientation), but the post presents this as a reference implementation, and the ladder's monotonicity claim depends on each fit attaining its group's optimum.

**5. `index.qmd:223` — "±5 pixels … is worth about two degrees" is wrong.** Simulating σ=5 px on the actual landmarks gives a nose-angle sd of 3.6° (frame 4) to 6.3° (frame 1). The conclusion (nowhere near 31°) survives; the quoted number does not.

**6. `index.qmd:482` — a Gaussian σ is labelled as a bound.** `rng.normal(0, 5, ...)` has RMS radial error 5√2 ≈ 7 px with ~32% of points beyond 5 px, but the prose says "±5 pixel jitter" and compares it directly to a hand-placement claim of "±5 pixels" stated as a bound. The direction is conservative, but the final conclusion *is* that comparison, so say so.

### Suggestions

**7. `index.qmd:36` / `src/prepare_photos.py:29` — orientation.** The committed JPEGs are 1512 × 2016 (portrait), so the originals were 3024 × 4032, not "4032 × 3024". Related: the `fig-matrix` title prints rows × cols = "2016x1512" under prose saying "1512 × 2016", so the figure appears to contradict the text.

**8. `src/shapes_analysis.R:70` — duplicated comment block** (lines 70–72 and 73–75), the first copy referring to `affine_gpa`, a function that no longer exists. Leftover from an earlier revision.

**9. `index.qmd:392` — "The ladder is monotone, as it has to be."** It is monotone on this data, but nothing in the code guarantees it: `fit_similarity` is non-optimal in its reflection branch (4), and `fit_projective` is a DLT seed plus local LM refinement with no global guarantee — its own docstring concedes the DLT can score worse than the nested affine fit. The prose asserts as a theorem something the implementation only happens to satisfy.

### Strengths

Numbers reproduce exactly across both pipelines; the Python/R cross-check is genuine rather than decorative. The affine-collapse trap and the DLT algebraic-error trap are documented as callouts rather than quietly patched. The post already declines to claim the sail billow — findings 1 and 6 sharpen that caution rather than overturning it.

PR #170, opened 2026-08-18
feat(posts): warp the kite photographs themselves, in R

project-delphi:
## Summary

Follow-up to #169. The R section of `posts/kite-shape-analysis/` was landmarks-only — `tpsgrid` bends an abstract mean, and nothing on the page let a reader see the kites in register. This sends the photographs through the same pipeline.

- **`src/warp_movie.R`** (new) transports each frame into one shared panel twice, by the best similarity map and by the best affine map, then morphs between consecutive frames: landmarks interpolated linearly, both photographs bent onto the intermediate shape by thin-plate spline, cross-dissolved.
- **`fig-r-warped-stack.png`** — as photographed / similarity-aligned / affine-aligned, five frames each. Carries the finding statically so the GIF is not the only copy of a result.
- **`fig-r-morph.gif`** — the two-panel animation, 0.41 MB (repo's existing GIFs are 0.18–0.41 MB).
- **`fig-r-plotshapes.png`** — `shapes::plotshapes` on raw vs `procGPA` configurations.

## The finding is a negative one, and the post says so

Measured on pixels the two panels are **indistinguishable**: per-pixel SD across the five aligned frames is 0.128 under similarity, 0.129 under affine. Pixel disagreement here is dominated by which colour bands a viewpoint exposes, by lighting, and by the streamers — which are in no correspondence at all.

What the affine map actually improves is landmark registration, 7.7 → 5.6 panel pixels (~28%, consistent with the 38% Procrustes SS reduction the post already reports). On a 220-pixel panel that is invisible, so the animation draws residual spikes at **three times life size**, captioned as such — `tpsgrid` has a `mag` argument for the same reason. This is written up as a result, not worked around: a real 38% improvement can look like nothing at all when painted onto pixels.

The streamers tear and smear in both warped rows. That is kept and explained inline — a TPS is pinned only where you pin it, and outside the hull of six sail landmarks it is extrapolating from nothing.

## Notes

- **Morpho is deliberately not used**, though it would have supplied `tps2d`. Its CRAN binary links `/Library/Frameworks/R.framework/.../libRblas.dylib`; this Homebrew R ships `libRlapack.dylib` and no `libRblas.dylib`, so it installs and then fails at `dlopen` (same for `RcppArmadillo`). The spline is written out longhand instead, which suits the post. Documented in `src/r-requirements.txt`.
- Housekeeping: `fig-landmarks.png` 1.33 MB → 0.69 MB, and `landmarks.csv` now written with LF so git stops reporting it modified on every checkout.

## Test plan

- [x] `Rscript src/warp_movie.R` completes; all three assets written
- [x] GIF frames extracted with `ffmpeg` and inspected — sail registers, residual spikes visibly shorter on the affine panel
- [x] Panel difference verified numerically before trusting the visual (7.67 → 5.55 px), and the ghosting check that showed *no* pixel-level difference is what drove the magnified-residual design
- [x] `landmarks.csv` byte-identical after CRLF strip; `shapes_analysis.R` still reports SS 0.11100 / 0.06839 / 38.4%
- [x] `quarto render posts/kite-shape-analysis/index.qmd` and full-site render clean; all 10 image paths resolve in `docs/`
- [x] `python3 scripts/check_posts.py` OK
- [x] `pre-commit` scoped to changed files: black, ruff, pyupgrade, add-trailing-comma all pass


project-delphi (commented):

`warp_movie.R` was re-run end-to-end against a scratch output dir and reproduces all three committed assets byte-for-byte (429594 / 72174 / 729298), so the figures match the script.

### What checks out

The TPS algebra is textbook and correct (`L = [[K,P],[P',0]]`, exact interpolation, coefficients sliced in the right order). `fit_similarity` is the correct reflection-allowed Procrustes: `svd(t(d0) %*% s0)` yields `u %*% t(v) = t(R_opt)`, applied as `pts %*% t(rot)`, so the rotation is not transposed; `scale = sum(d)/sum(s0^2)` is the standard optimum. `fit_affine` is the right least-squares map in the same `apply_map` convention, and `invert_map` correctly gives `rbind(-t %*% A^-1, A^-1)`. The inverse-warp direction is right, the bilinear sampler's row/col ordering matches `expand.grid(y=…, x=…)` filling `matrix(v, BOX, BOX)`, and `rasterImage` orientation is consistent in both the reversed-y morph panels and the normal-y stack.

Similarity and affine are **not** swapped in either figure, and no transport carries a reflection (all dets 0.33–0.79). Landmark residual RMS at BOX=220 is 7.095 → 5.107 panel px — exactly the quoted 7.0 → 5.1 and 28.0%. At BOX=240 they would be 7.74/5.57, confirming the stale-number fix in daf2219. `make check-posts` passes and `docs/` is in sync.

### Findings

**[High] `index.qmd:750` — the 0.128 / 0.129 per-pixel SD figures don't reproduce.** Mean over pixels of the SD across the five aligned frames at BOX=220 gives **0.0724 (similarity) / 0.0738 (affine)** — about 1.75x smaller than published. Every plausible alternative definition was tried and none lands on 0.128: greyscale-first (0.0691/0.0707), SD over all 15 frame×channel values (0.0776/0.0790), RMS of per-pixel SDs (0.0852/0.0860), pooled SD (0.0943/0.0958), mean |deviation| (0.0536/0.0543), spatial SD within each image (0.0874/0.0889), mean pairwise RMS difference (0.1188/0.1203, closest), and the same on the rendered panel PNGs (0.0486/0.0490). It is also identical at BOX=240, so daf2219's claim that this pair "was re-measured at the final size" cannot be what happened. The qualitative point survives (affine is marginally *higher* either way) but the magnitude is wrong.

Compounding it: **`warp_movie.R` computes neither the SD nor the landmark residuals**, so neither published statistic is reproducible from the committed script. Both should be `cat()`ed at the end of the run so the prose can be checked against it.

**[Low] `warp_movie.R:224`** — "a few pixels on a 240-pixel panel" survives the BOX 240 → 220 resize that daf2219 otherwise corrected.

**[Low] `warp_movie.R:156`** — the "as photographed" baseline is neither unaligned nor native scale. `crop_raw` centres on `colMeans(X[,,i])`, removing translation per frame, and samples a 300-native-pixel window onto a 220-pixel panel (0.73x). The comment says "No alignment of any kind" and `index.qmd:717` says "a fixed window at native scale and orientation", while `index.qmd:723` credits the top-to-middle step with removing "position, size and rotation" — but position is already gone in the top row.

**[Low] `warp_movie.R:164`** — figure layout hardcodes n=5 and k=6 while both are derived: `par(mfrow=c(3,5))`, `joinline=c(1:6,1)` (186/188), and the six-entry `DOT` palette (212). A sixth photograph would silently wrap the grid and recycle colours rather than fail. Line 211's `hull <- c(1:k, 1)` already does it the general way.

**[Low] `src/r-requirements.txt:20`** — `colorRamps` is pinned and in the install line but unused; `shapes` 1.2.6 imports only `minpack.lm`, `scatterplot3d`, `rgl`, `MASS`. An unused pin misleads in a file whose job is recording what the post needs.

PR #171, opened 2026-08-18
feat(posts): add the statistical jackknife post

project-delphi:
## Summary

A new post on the jackknife, weighted toward its four worked examples rather than its derivations.

- **Math moved out of the way.** Quenouille's cancellation and the von Mises influence-function expansion now live in collapsible callouts. Only the three formulas the reader computes with — the estimator, the pseudovalue, the variance — stay inline.
- **Examples explained properly.** Each of the four gets a plain-language warm-up, a prose paragraph situating its simulated data (generating process, what real situation it stands in for, why simulating beats a real dataset there, what a wrong decision costs), and a walkthrough quoting the numbers it actually produced. The four interactive labs each get a paragraph on what the controls do and what to watch for.

**Three examples did not support the claims made about them**, so the code changed too:

| | Was | Now |
|---|---|---|
| Ratio bias figure | `y = 2.5*x + noise` makes `Cov(X̄,Ȳ) = R·Var(X̄)`, cancelling the leading bias identically — measured bias at n=5 was **+0.0008**, against prose claiming "more than 0.12". Both curves flat on zero. | Y drawn independently of X. Plug-in bias **+0.103** at n=5 decaying as ≈0.5/n; jackknife within 0.007 of zero throughout. |
| Fraud figure | Leverage ranked the seven zero-spend bots (0.01404) **above** the ringleader (0.01330, rank 8 of 16). The chart showed five identical blue bot bars and no ringleader, contradicting "an unmistakable outlier". | Each account charged the credits it *caused* — its own $20 plus $20 per referee — which is the referrer-and-referee bonus the section text already described. Ringleader **0.111** vs 0.014, ~8× the next account. |
| Biomarker figure | Artifact spike of 14.5 gave LINC009 a naive mean of **0.60** against CDK1's 3.06, so "naive fold-change ranks both as strong hits" was false. | Spike raised to 90 (realistic for an RNA-seq count artifact). Naive means **3.12** vs **3.06** — the artifact outranks the real target — while stability reads **1.04** vs **28.19**. |

The survival regression reported a paid-search effect of −55.2 days as exact when the simulated truth was −28.9 and the draw was ~1 SD out. It now runs at n=400 and prints a robust sandwich SE and the simulated truth beside every coefficient; all three 95% intervals cover.

`widgets.js` carried the same bug as the fraud figure (`promo_cost = bonus` for every account), so Lab 2 put bot spokes above the hub in the reader's browser. Same referral attribution applied, driven by the previously unused `referrals` field. Also replaced five raw-LaTeX strings that rendered literally, since MathJax does not process injected DOM.

Separately: prose rules consolidated into `STYLE.md` (CLAUDE.md keeps the pointer plus data provenance, frontmatter, assets), and the code/figure seam rule dropped in 044c805 restored — STYLE.md's nearest rule governs warm-up length, not seams, so a distinct rule had been lost rather than de-duplicated.

`docs/index.html` also picks up a reading-time refresh on the kite post (32→33 min); that drift predates this branch and the render corrects it.

## Test plan

- [x] `QUARTO_PYTHON=.venv-jackknife/bin/python quarto render posts/statistical-jackknife/index.qmd` clean
- [x] `make check-posts` passes — freeze matches source, `jackknife-blog` added to `kernels-stub`
- [x] Every number quoted in prose read back off the rendered HTML and figures
- [x] All three figures inspected: bias curve decays, ringleader red and top, biomarker artifact outranks the target on the naive score
- [x] All four widgets exercised in Chrome — mount, respond, no console errors
- [x] Lab 2 slider swept 0/8/15 bots: hub tops the table at every setting, lead falls from ~9× to ~2× as spokes go to zero
- [x] Both collapsible callouts open and render their MathJax


project-delphi (commented):
## Automated review (`/code-review`, medium)

Reviewed at `main...HEAD`. Every Python number in the prose was re-verified against the committed output, and the four widgets' math was re-implemented independently to check the claims made about them. `make check-posts` passes; `docs/` is rendered and committed alongside source; `jackknife-blog` is registered in `kernels-stub`.

### Critical

- **`widgets.js` — Lab 4's median caption stated the inverse of what the widget draws.** With `sample = [1.2 … 9.8]` (n=11, median 3.8), deleting the *central* point moves the median by exactly `0.00` and deleting any *non*-central point moves it by exactly `±0.20`. The caption claimed the opposite, and this is the one lab whose entire purpose is the non-smoothness lesson. It also contradicted the post body, which was correct. **Fixed** — caption now states the real behaviour, including that extremeness is irrelevant.

### Important

- **Lab 3 promised a ranking flip that did not occur.** The artifact was hardcoded at `diffs[14] = 12.5`, roughly an order of magnitude too small: LINC009 sat at naive rank #4 both before and after, and never entered the stability top-5 in either state, so the button changed one cell. **Fixed** — injection now reads `outlier_patient`/`outlier_val` from the data file. Verified: with artifacts injected, LINC009 (5.11) and FAM83A (4.74) take the top two naive slots above CDK1 (4.69) and MYC (4.41), while the stability top-5 (CDK1/EGFR/BRCA1/TP53/MYC, 38.7–48.6) is unchanged and both artifacts score 1.02.
- **Lab 3 ignored its own data file's `outlier_patient`/`outlier_val`,** so FAM83A — declared `"type": "brittle"`, "driven entirely by Patient #3" — rendered as perfectly stable. **Fixed** by the same change.
- **Lab 1 was described as supporting interactions it never had.** The prose said "drag a point, add one, delete one"; only `mousedown`/`touchstart` are registered. **Fixed** — prose now describes dragging only.
- **`widget-data/smooth_vs_jagged.json` was dead payload contradicting the widget.** `initWidget4` takes no data argument, so its hardcoded 11 points were always used while the file held 15 with a different maximum. **Fixed** by deleting the file and its payload entry; the hardcoded sample is the one the prose describes and gives the cleaner exactly-zero centre deletion.
- **Lab 2 mutated the shared account objects it was passed.** Behaviour was idempotent so nothing broke, but the file's hub `referrals` was silently overwritten. **Fixed** — attribution now maps to copies; the seven never-read `syndicate_spoke` rows were dropped from the data file and the hub's referral count set to the slider default.

### Suggestions

- **Unused imports** (`sys`, `matplotlib as mpl`, `scipy.stats`) in the setup cell. **Fixed.**

### Strengths

Nothing wrong found in the Python, the frontmatter, the `_freeze` record, the `Makefile` kernel addition, or the CLAUDE.md → STYLE.md prose-rule move — that split is clean and leaves no dangling reference. All prose numbers (+0.1034 plug-in bias, 0.1106 ringleader leverage, 204.2 RMST, the three "covered: yes" rows, 28.19 vs 1.04 stability) match the committed output exactly, and Lab 2's "nine times at eight spokes / barely twice at zero" reproduces at 8.87 and 2.15.

All findings addressed in `7612b85`. Every Python output is byte-identical after the fixes.

project-delphi (commented):
## Re-review after the critical fix (`/code-review`, medium)

Second pass over `widgets.js`, `index.qmd`, and `widget-data/`. **No Critical findings** — the Lab 4 caption fix holds, and every Python number still matches the frozen output. The remaining findings were all prose-versus-widget mismatches of the same family, and are fixed in `f454acc`.

### High

- **Lab 1's instruction was falsified by its own widget.** The statistic is an OLS *slope*, so influence follows x-leverage, not vertical distance from the fit. Verified against `toy.json`: point #12 (x=5.0) has the largest residual in the sample by far (+3.33 against a range of −0.05 to −0.60) and an influence of only −0.05; dragging it from y=9.5 to y=0.5 moves its own bar to +0.08 while others swing by 0.3–0.4. The promised "its own bar grows while the rest barely stir" only happens for an *end* point (#1 at y=9.0 gives −3.90 against a next-largest 1.30). **Fixed** — the reader is sent to an end point, and the leverage-vs-residual gap is now the explicit lesson rather than a contradiction.

### Important

- **Lab 3's naive panel ranked on signed `meanLFC`** while the stability panel ranked on a magnitude. BRCA1 (−3.6) and TP53 (−3.5), planted as genuine down-regulated targets, could never place, and LINC009 (0.10) and ACTB (0.10) — pure noise — held ranks 4 and 5. **Fixed** by sorting on `Math.abs(meanLFC)`; the baseline naive top-5 is now CDK1/MYC/EGFR/BRCA1/TP53.
- **The Lab 3 button read "into LINC009" but injected FAM83A too**, and highlighted by hardcoded gene name rather than by what was actually injected. **Fixed** — label names both genes, and the ⚠️ highlight follows an `injected` flag. Verified both are flagged after the click.
- **Lab 2 never displayed leverage**, though the prose asks the reader to compare it at two slider settings, and its readout was gated on `botSpokes > 0` — hiding it at zero spokes, one of the two endpoints named. **Fixed** — a Leverage column, and a banner that stays at zero spokes. Verified it now reports 8.9× at eight spokes and 2.2× at zero, matching the prose.
- **Lab 1's SVGs were fixed-width with no `max-width`**, overflowing a phone viewport. **Fixed** — `max-width:100%; height:auto`, and the drag handler now rescales client pixels into viewBox units, since it previously assumed they were the same. Verified at a 320px container: SVG scales (k=1.19) and a drag still lands within one viewBox unit of target, with zero document overflow.

### Suggestions

- **Freeze drift**: Quarto hashes `index.qmd` alone, so editing `widgets.js` or `widget-data/*.json` leaves `freeze: auto` serving the old bundle while `make check-posts` still reports OK. **Addressed** with a note in the injection cell telling the next editor to re-render explicitly.
- **Unguarded outlier index** — `outlier_patient >= n` would produce a sparse array with wrong means and no error. **Fixed** with a bounds check.
- **Misleading comment** — `Math.sin(i * 3 + gene.name.length)` was labelled "Normal dist approx"; it is a deterministic sinusoid. **Comment corrected.**

### Strengths

Nothing wrong found in the Python, frontmatter, `_freeze` record, `Makefile` kernel addition, or the CLAUDE.md → STYLE.md split. All prose numbers verified against the committed output; Lab 4's ±0.20 / 0.00 median claim is correct for its sample; `make check-posts` passes.

Every Python output is byte-identical across all three fix commits — only widget behaviour and prose changed.

